PR #7, opened 2015-02-20
Define Performance interface

### LINKED ISSUE #6: Expose performance interface to worker
ServiceWorker needs access to Performance interface. To enable this, I think we need to do a bit of spec shuffling and cleanup:

- Move definition of Performance interface from NavTiming ([sec 4.2](http://w3c.github.io/navigation-timing/#sec-performance-interface)) into Performance Timeline spec, and update Nav Timing language to "participate" in the timeline.
- In Perf Timeline expose Performance interface to [Window,Worker].
- Cleanup Nav/Resource/User Timing and remove references to 'window.performance'. 

@plehegar any objections?

igrigorik:
- Moves definition of Performance interface from NavTiming into Performance Timeline (NavTiming update will be handled in a separate changeset).
- Performance interface is exposed to [Window, Worker].
- Other editorial cleanup.

@plehegar: can you take a quick pass?

Closes #6.

### DIFF
--- a/index.html
+++ b/index.html
@@ -126,15 +126,13 @@
           <dt>readonly attribute DOMString name</dt>
           <dd>The attribute MUST return the identifier for this <a>PerformanceEntry</a> object. This identifier does not have to be unique.</dd>
           <dt>readonly attribute DOMString entryType</dt>
-          <dd>This attribute MUST return a <a href="http://www.w3.org/TR/WebIDL/#idl-DOMString">DOMString</a> that describes the type of the interface represented by this <a>PerformanceEntry</a> object.</dd>
+          <dd>This attribute MUST return a <a href="http://www.w3.org/TR/WebIDL/#idl-DOMString">DOMString</a> that describes the <a href="http://www.w3.org/wiki/Web_Performance/EntryType">type of the interface</a> represented by this <a>PerformanceEntry</a> object.</dd>
           <dt>readonly attribute DOMString startTime</dt>
           <dd>The attribute MUST return a <a href="http://www.w3.org/TR/hr-time/#domhighrestimestamp">DOMHighResTimeStamp</a> that contains the time value of the first recorded timestamp of this performance metric.</dd>
           <dt>readonly attribute DOMString duration</dt>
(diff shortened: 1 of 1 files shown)


plehegar:
I believe the following is incorrect:

[Exposed=Window,Worker]
interface Performance : EventTarget { ... }

It exposes all the methods/attributes of the Performance interface directly on Window and Worker, while we only want to expose a performance attribute on Window and Worker.

You need to do something like:
interface Performance : EventTarget { ... }

[Exposed=Window,Worker]
interface PerformanceExtension {
    [Replaceable]
    readonly    attribute Performance performance;
};

plehegar:
Btw, I think it would read better if the performance interface was define first, ie switch the appearance of performanceEntry and performance definitions in the spec

igrigorik:
@plehegar I'll defer to you on the IDL.. still trying to find my way around it. 

That said, I've read the [definition of `Exposed`](http://heycam.github.io/webidl/#Exposed) as sufficient to what we're looking for: "it indicates that the interface, interface member or dictionary constructor is exposed on a particular set of global interfaces, rather than the default of being exposed only on the primary global interface." What does [Replaceable](http://heycam.github.io/webidl/#Replaceable) give us on top? 

FWIW, [NetInfo](http://w3c.github.io/netinfo/#the-networkinformation-interface) doesn't seem to define Replaceable either.. but perhaps it should? /cc @marcoscaceres

plehegar:
@igrigorik I'll do a pass at the PR and modify the WebIDL. With your current interface, you'd get window.getEntries instead of windows.performance.getEntries.
Replaceable comes from the navigation timing spec itself and allow the webapp to override the performance attribute. It was our way to make sure we wouldn't break pages that were using window.performance at the beginning. We may be able to remove it in the future but it should be a separate issue imho.

igrigorik:
Ah, that makes sense. Thanks, I'll leave this in your capable hands :)

igrigorik:
@plehegar anything I can do to help land this? :)

plehegar:
Here is what I think needs to be done. First, I was incorrect on Exposed. You were right to add on the Performance interface. It exposes the *interface* to the global object, not its attributes directly. I added to PerformanceEntry as well. Finally, I added support for the Performance attribute directly on Window and GlobalWorkerScope.

igrigorik:
lgtm, modulo the small and unrelated http:// nitpick (comment on the commit).

PR #8, opened 2015-03-18
(WIP) PerformanceObserver

igrigorik:
**This is a work in progress.**

See [proposal doc](https://docs.google.com/document/d/1fXtxtPC1Gg4PeLXI_axj6AvMTznf9X5lrj5HTyR3r3w/edit#) and the [public-web-perf thread](https://lists.w3.org/Archives/Public/public-web-perf/2015Feb/0047.html) for full context.

### DIFF
--- a/index.html
+++ b/index.html
@@ -164,16 +164,68 @@
         WorkerGlobalScope implements GlobalPerformance;
         </pre>
       </section>
+
+    <section>
+      <h2>The <dfn>PerformanceObserver</dfn> interface</h2>
+
+      <p>A <a>PerformanceObserver</a> object can be used to observe the <a>Performance Timeline</a> and be notified of new performance metrics as they are recorded by the user agent.</p>
+
+      <dl title='callback PerformanceObserverCallback = void (PerformanceObserverTimeline eventList, PerformanceObserver observer)' class='idl'></dl>
+
+      <dl title='[Constructor(PerformanceObserverCallback callback), Exposed=(Window,Worker)] interface PerformanceObserver' class='idl'>
+
+        <dt>void observe(PerformanceObserverInit options)</dt>
+        <dd>This method instructs the user agent to <dfn>register the observer</dfn> (i.e. observe the Performance Timeline), and report any new performance entries based on the criteria given by <a>options</a>.
+          <dl title='dictionary PerformanceObserverInit' class='idl'>
+            <dt>sequence&lt;DOMString&gt; typeFilter</dt>
+            <dd>
+              <p>A list of valid <a href="#widl-PerformanceEntry-entryType">entryType names</a> if not all entry types need to be observed. The list must not be empty and types not recognized by the user agent must be ignored.</p>
+
+              <p class="note">To keep the performance overhead to minimum the application should only subscribe to event types that it is interested in, and disconnect the observer once it no longer needs the performance data. Filtering by name is not supported, as it would implicitly require a subscription for all event types &mdash; this is possible, but discouraged, as it will generate a significant volume of events.</p>
+            </dd>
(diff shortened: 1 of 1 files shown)


igrigorik:
First and **_very rough_** run (661992f) at the PerformanceObserver interface. There is a lot of stuff missing, and the IDL is probably wrong, but should be enough to start the conversation: [see rawgit preview](https://rawgit.com/w3c/performance-timeline/661992ffe412f031d6cb83c9509a83a5d35c6c52/index.html#the-performanceobserver-interface). 

Note that above draft proposes a slightly different model than what we've discussed on the ML: 
* Existing `Performance` interface operates on global timeline
* I'm proposing that each `PerformanceObserver` gets its own ("local timeline") `PerformanceObserverTimeline` which records only the events that match its filter types (passed by `PerformanceObserverInit` dictionary), and only while the observer is active. This is heavily inspired by [MutationObserver interface](http://www.w3.org/TR/dom/#mutation-observers)...
  * PerformanceObserverTimeline provides the same `getEntries`, `getEntriesByType`, `getEntriesByName` methods to access the timeline as current `Performance` interface.
  * Instead of `takeRecords()` method exposed by MutationObserver, which returns all of the events in the "local timeline", we provide `clear()` which clears the timeline without returning any records. 

Conceptually, I think the above model has a number of benefits:
* The global vs local timeline is easy to explain and understand
* Each observer has local timeline which is scoped based on its subscription preferences
* Each local timeline exposes same methods as global timeline, plus clear() which flushes all records
* PerformanceObserverTimeline can be implemented efficiently by the UA:
  * UA can continue appending "lazy records" and fire notifications to the subscriber without necessarily creating the underlying performance objects in JS-land; objects only have to be materialized when accessed via one of getEntries* methods.
 * Application code can clear the local timeline to get efficient access to deltas and/or choose to buffer events until its ready to process them; the buffering is done automatically by the UA and the application does not need to define own logic to log every new notification.

Thoughts, feedback?

natduca:
How about doing a PR that adds what we agreed upon on the ML, then do followup PRs that add the individual changes that you propose here?

E.g. you have a pr for a clear method, and one for the addition of the PerformanceObserverTimeline object.

I personally am suportive of the clear addition, and unsupportive of the timeline object, but I think that its better to address each in a separate change discussion

igrigorik:
@natduca apologies about the confusion. The "PerformanceObserverTimeline" is actually just a different name for "LazyPerformanceEntryList" from our [earlier doc](https://docs.google.com/document/d/1fXtxtPC1Gg4PeLXI_axj6AvMTznf9X5lrj5HTyR3r3w/edit). If you ignore `clear()` and different name, then 661992f is exactly what was proposed.

Followed up on public-web-perf with the larger context: https://lists.w3.org/Archives/Public/public-web-perf/2015Mar/0031.html -- I'll wait for feedback on that, which is probably where I should have started. :)

natduca:
Thanks for the clarification.

I think though that PerformanceObserverTimeline should remain titled MumblePerformanceEntryList, because thats closer to what it is. For some value of Mumble. Its not a timeline, you're not supposed to interact with it like the timeline object. Its a list of events, thats it.

It seemed we had buyoff on the doc as it was... I really think we should do the renamings and everything as individual steps. Unless I missed something on email somewhere?

Similarly, the clear should be a separate change proposal somewhere: PR/mail/etc... the MumblePerformanceEntryList is meant to be immutable: that lets the implementation be lean. By adding clear, it becomes mutable and that raises a bunch of questions I'm not sure we really want to get into...

igrigorik:
(closing, will open new pull with earlier version to avoid polluting history)

PR #10, opened 2015-04-02
PerformanceObserver

igrigorik:
**Note: this is a WIP pull... expect more commits to land as we iterate on this.**

Preview: https://rawgit.com/w3c/performance-timeline/performanceobserver/index.html

---
* Defines [PerformanceObserver interface](https://rawgit.com/w3c/performance-timeline/performanceobserver/index.html#the-performanceobserver-interface) 
  * Inspired by [Mutation Observers](http://www.w3.org/TR/dom/#mutation-observers) and can be used to subscribe to / observe new performance events delivered into the Performance Timeline. 
   * Creating a subscription requires that you provide a list of entryTypes that you want to listen for.

* Defines [LazyPerformanceEntryList interface](https://rawgit.com/w3c/performance-timeline/performanceobserver/index.html#the-lazyperformanceentrylist-interface) 
  * Provides getEntries, getEntriesByType, getEntriesByName methods

---

Questions:
* [ ] Sanity check: do we really need getEntriesBy* on LazyPerformanceEntryList? If the application is only interested in particular type it can create a dedicated Observer and filter for just that type. Would it make sense to simplify this down to getEntries()? 
* [ ] Define processing steps. How do we specify this, ala [Mutation Observers with microtasks](http://www.w3.org/TR/dom/#queuing-a-mutation-record), etc?
  * Can we define "queuing a performance entry record" that is shared by timeline and observer?
  * When is entry queued for delivery? For observer, once all of its values have been populated? E.g. when is a resource or navigation entry queued. How/when do we queue the record into the appropriate buffer?

/cc @bzbarsky @natduca @plehegar @mpb @toddreifsteck @paulirish

### DIFF
--- a/index.html
+++ b/index.html
@@ -109,7 +109,7 @@
     });
 
     // subscribe to Frame-Timing and User-Timing events
-    observer.observe({eventTypes: ['renderer', 'composite', 'mark', 'measure']});
+    observer.observe({eventTypes: ['render', 'composite', 'mark', 'measure']});
     observer.disconnect();
 
     // processing logic called by application code
--- a/index.html
+++ b/index.html
@@ -189,7 +189,7 @@
     <section>
       <h2>The <dfn>Performance Observer</dfn> interface</h2>
 
-      <p>The <a>PerformanceObserver</a> interface can be used to observe the <a>Performance Timeline</a> and be notified of new performance entries as they are recorded by the user agent.</p>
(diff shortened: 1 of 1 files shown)


toddreifsteck:
LGTM

igrigorik:
As per discussion on the call today... merging, we'll iterate on the rest in separate issues:

- we can revisit getEntries* once we land #9 
- processing steps will be done as separate pull

Issue #3, opened 2014-10-28
PerformanceTimeline is async


plehegar:
The spec should make it clear it's async.

Issue #1, opened 2014-10-16
Notification for new performance entries


igrigorik:
From: http://lists.w3.org/Archives/Public/public-web-perf/2014Feb/0052.html 

@josh: Navigation and user timings have been useful to gather and analysis on
the server side. Though, its a bit tricky to build tools that just
take that data and upload it to a server. Theres no notification when
new performance entries have been recorded. This requires invasive
code to recheck the performance entries buffer.

```
    var seen = 0;
    function reportPerformance() {
      var entries = performance.getEntries();
      var newEntries = entries.slice(seen);
      seen = entries.length;
      if (newEntries.length) {
        navigator.sendBeacon("/stats",
          JSON.stringify(newEntries));
      }
    });

    // gather all entries from initial load
    window.addEventListener("load", function() {
      reportPerformance();
    });

    // report script timing after async script loads
    var script = document.createElement("script");
    script.src = "/async-script.js";
    script.onload = reportPerformance;
    document.head.appendChild(script);

    // report user timing
    window.performance.mark("mark_fully_loaded");
    reportPerformance();
```

And knowing when its safe to read performance.timing stats has a small
edge case.

```
    window.addEventListener("load", function() {
      performance.timing.loadEventEnd; // 0
      setTimeout(function() {
        performance.timing.loadEventEnd; // 123
      }, 0);
    });
```

It would be helpful to receive an event or callback when the browser
knows there are new performance entries available. This should cover
any resource and user timings. Anything that will add new entries to
`performance.getEntries()`. The notification should not even be
synchronous. Batching and throttling the notification would be up to
the browser.

As of the NavigationTiming2 spec, the Performance interface is already
an EventTarget. Heres an example of using a fictitious event name with
to log client performance timings back to a server backend.

```
    performance.addEventListener("entries", function(event) {
      navigator.sendBeacon("/stats",
        JSON.stringify(event.newEntries));
    });
```

And maybe even add an event or promise for navigation timing readiness.

```
    performance.timing.ready().then(function() {
      navigator.sendBeacon("/stats",
        JSON.stringify(performance.timing));
    });
```

---

> @jainarvind: Seems like a good idea. - [source](http://lists.w3.org/Archives/Public/public-web-perf/2014Feb/0055.html)

[...]

PR #11, opened 2015-04-30
Extend getEntries to allow attribute filtering

igrigorik:
Preview: https://rawgit.com/w3c/performance-timeline/getEntries-filter/index.html#methods

For background and context, see: https://github.com/w3c/performance-timeline/pull/9#issuecomment-89042906. 

- getEntries() returns copy of all items (as before)
- (new) getEntries(optional dictionary):
-- allows attribute queries against PerformanceEntry objects
- getEntriesByType(type) -> getEntries({'entryType': type})
- getEntriesByName(name) -> getEntries({'name': name})
- getEntriesByName(name, type) -> getEntries({'name': name, 'entryType':
  type})

### DIFF
--- a/index.html
+++ b/index.html
@@ -173,15 +173,41 @@
 
         <p>The <a>Performance</a> interface hosts performance related attributes and methods used to retrieve the performance metric data from the <a>Performance Timeline</a>.</p>
 
+        <dl title='dictionary EntryProps' class='idl'>
+          <dt>DOMString? name = null</dt>
+          <dd><a href="#widl-PerformanceEntry-name">name</a> of `PerformanceEntry` object.</dd>
+          <dt>DOMString? entryType = null</dt>
+          <dd><a href="#widl-PerformanceEntry-type">entryType</a> of `PerformanceEntry` object.</dd>
+          <dt>DOMString? initiatorType = null</dt>
+          <dd><a href="http://www.w3.org/TR/resource-timing/#widl-PerformanceResourceTiming-initiatorType">initiatorType</a> of `PerformanceEntry` object.</dd>
+        </dl>
+
         <dl title='[Exposed=(Window,Worker)] interface Performance : EventTarget' class='idl'>
-          <dt>PerformanceEntryList getEntries()</dt>
-          <dd>This method returns a <a>PerformanceEntryList</a> object that contains a copy of all <a>PerformanceEntry</a> objects in chronological order with respect to <a href="#widl-PerformanceEntry-startTime">startTime</a>.</dd>
+          <dt>PerformanceEntryList getEntries(optional EntryProps filter)</dt>
+          <dd>
+            <p>This method returns a <a>PerformanceEntryList</a> object that contains a copy of <a>PerformanceEntry</a> objects, sorted in chronological order with respect to <a href="#widl-PerformanceEntry-startTime">startTime</a>, that match the following criteria:</p>
+
+            <ol>
+              <li>Let the <dfn>list of entry objects</dfn> be the empty <a>PerformanceEntryList</a>.</li>
+              <li>Let the <dfn>set of filter properties</dfn> be the non-null <a href="http://www.w3.org/TR/WebIDL/#dfn-dictionary-member">dictionary members</a> of `filter`, or the empty set if `filter` is omitted.</li>
+              <li>For each available <a>PerformanceEntry</a> object (<dfn>entryObject</dfn>), in chronological order with respect to <a href="#widl-PerformanceEntry-startTime">startTime</a>:</li>
+                <ol>
+                <li>For each _key_ and _value_ pair in <a>set of filter properties</a>:</li>
+                  <ol>
+                  <li>If the <a>entryObject</a> does not contain an attribute whose name matches _key_, go to next <a>entryObject</a>.</li>
+                  <li>Otherwise, if the <a>entryObject</a> contains an attribute whose name matches _key_, and its value does not match _value_, go to next <a>entryObject</a>.</li>
+                  </ol>
+                <li>Add <a>entryObject</a> to the <a>list of entry objects</a>.</li>
+                </ol>
+              <li>Return the <a>list of entry objects</a>.</li>
+            </ol>
+          </dd>
 
           <dt>PerformanceEntryList getEntriesByType(DOMString entryType)</dt>
-          <dd>This method returns a <a>PerformanceEntryList</a> object that contains a copy of all <a>PerformanceEntry</a> objects, in chronological order with respect to <a href="#widl-PerformanceEntry-startTime">startTime</a>, that have the same value for the <a href="#widl-PerformanceEntry-entryType">entryType</a> attribute of <a>PerformanceEntry</a> as the <a href="#widl-PerformanceEntry-entryType">entryType</a> parameter. If no such <a>PerformanceEntry</a> objects exist, the <a>PerformanceEntryList</a> must be empty.</dd>
+          <dd>This method returns a <a>PerformanceEntryList</a> object returned by `getEntries({'entryType': entryType})`.</dd>
 
           <dt>PerformanceEntryList getEntriesByName(DOMString name, optional DOMString entryType)</dt>
-          <dd>This method returns a <a>PerformanceEntryList</a> object that contains a copy of <a>PerformanceEntry</a> objects, in chronological order with respect to <a href="#widl-PerformanceEntry-startTime">startTime</a>, that have the same value for the <a href="#widl-PerformanceEntry-name">name</a> attribute of <a>PerformanceEntry</a> as the <a href="#widl-PerformanceEntry-name">name</a> parameter and, if specified, have the same value for the <a href="#widl-PerformanceEntry-entryType">entryType</a> attribute of <a>PerformanceEntry</a> as the <a href="#widl-PerformanceEntry-entryType">entryType</a> parameter. If no such <a>PerformanceEntry</a> objects exist, the <a>PerformanceEntryList</a> must be empty.</dd>
+          <dd>This method returns a <a>PerformanceEntryList</a> object returned by `getEntries({'name': name})` if optional `entryType` is omitted, and `getEntries({'name': name, 'entryType': entryType})` otherwise.</dd>
         </dl>
 
         <dl class='idl' title='typedef sequence<PerformanceEntry> PerformanceEntryList'></dl>
@@ -236,8 +262,9 @@
 
         <p>The <a>PerformanceObserverEntryList</a> interface provides the same `getEntries`, `getEntriesByType`, `getEntriesByName` methods as the <a>Performance</a> interface, except that <a>PerformanceObserverEntryList</a> operates on the observed emitted list of events instead of the global timeline.</p>
 
+
         <dl title='[Exposed=(Window,Worker)] interface PerformanceObserverEntryList' class='idl'>
-          <dt>PerformanceEntryList getEntries()</dt>
+          <dt>PerformanceEntryList getEntries(optional dictionary entryProps)</dt>
           <dd>See <a href="#widl-Performance-getEntries-PerformanceEntryList">performance.getEntries</a>.</dd>
           <dt>PerformanceEntryList getEntriesByType(DOMString entryType)</dt>
           <dd>See <a href="#widl-Performance-getEntriesByType-PerformanceEntryList">performance.getEntriesByType</a>.</dd>
--- a/index.html
+++ b/index.html
@@ -185,7 +185,7 @@
         <dl title='[Exposed=(Window,Worker)] interface Performance : EventTarget' class='idl'>
           <dt>PerformanceEntryList getEntries(optional EntryProps filter)</dt>
           <dd>
-            <p>This method returns a <a>PerformanceEntryList</a> object that contains a copy of <a>PerformanceEntry</a> objects, sorted in chronological order with respect to <a href="#widl-PerformanceEntry-startTime">startTime</a>, that match the following criteria:</p>
+            <p>This method returns a <a>PerformanceEntryList</a> object that contains a list of <a>PerformanceEntry</a> objects, sorted in chronological order with respect to <a href="#widl-PerformanceEntry-startTime">startTime</a>, that match the following criteria:</p>
 
             <ol>
               <li>Let the <dfn>list of entry objects</dfn> be the empty <a>PerformanceEntryList</a>.</li>
--- a/index.html
+++ b/index.html
@@ -189,13 +189,13 @@
 
             <ol>
               <li>Let the <dfn>list of entry objects</dfn> be the empty <a>PerformanceEntryList</a>.</li>
-              <li>Let the <dfn>set of filter properties</dfn> be the <a href="http://www.w3.org/TR/WebIDL/#dfn-dictionary-member">dictionary members</a> of `filter`, or the empty set if `filter` is omitted.</li>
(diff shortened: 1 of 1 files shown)


igrigorik:
@plehegar @toddreifsteck @yoavweiss PTAL, hopefully it doesn't look too crazy. Also, if you have any suggestions on IDL or better algorithm structure, let me know!

igrigorik:
@plehegar added missing "otherwise" and simplified the algorithm a bit in 24932b9.. The missing and empty cases should fall our directly out of it - i.e. missing and {} return the same set.

index.html
```diff
@@ -162,14 +162,31 @@
...
+
+            <ol>
+              <li>Let the <dfn>set of entry objects</dfn> be the empty <a>PerformanceEntryList</a>.</li>
+              <li>Let the <dfn>set of filter properties</dfn> be the value of `entryProps`, or the empty dictionary if `entryProps` is omitted.</li>
+              <li>For each available <a>PerformanceEntry</a> object (<dfn>entryObject</dfn>), in chronological order with respect to <a href="#widl-PerformanceEntry-startTime">startTime</a>:</li>
+                <ol>
+                <li>For each _key_ and _value_ pair in <a>set of filter properties</a>:</li>
+                  <ol>
+                  <li>If the <a>entryObject</a> does not contain an attribute who's name matches _key_, go to next <a>entryObject</a>.</li>
+                  <li>Otherwise, if the <a>entryObject</a> contains an attribute who's name matches _key_, and its value does not match _value_, go to next <a>entryObject</a>.</li>
```
lawnsea: s/who's/whose/ on this line and the previous

index.html
```diff
@@ -162,14 +162,31 @@
...
 
         <dl title='[Exposed=(Window,Worker)] interface Performance : EventTarget' class='idl'>
-          <dt>PerformanceEntryList getEntries()</dt>
-          <dd>This method returns a <a>PerformanceEntryList</a> object that contains a copy of all <a>PerformanceEntry</a> objects in chronological order with respect to <a href="#widl-PerformanceEntry-startTime">startTime</a>.</dd>
+          <dt>PerformanceEntryList getEntries(optional dictionary entryProps)</dt>
+          <dd>
+            <p>This method returns a <a>PerformanceEntryList</a> object that contains a copy of <a>PerformanceEntry</a> objects, sorted in chronological order with respect to <a href="#widl-PerformanceEntry-startTime">startTime</a>, that match the following criteria:</p>
+
+            <ol>
+              <li>Let the <dfn>set of entry objects</dfn> be the empty <a>PerformanceEntryList</a>.</li>
```
lawnsea: Given that the `PerformanceEntryList` is ordered, "set" doesn't quite seem the right word here. Perhaps "list" instead?

index.html
igrigorik: good catch, thanks!

igrigorik:
@lawnsea thanks, should be fixed now.

toddreifsteck:
I’m reviewing with a few folks internally. Will get back to you by 5/7 or 5/8.

index.html
```diff
@@ -162,14 +162,31 @@
...
+
+            <ol>
+              <li>Let the <dfn>list of entry objects</dfn> be the empty <a>PerformanceEntryList</a>.</li>
+              <li>Let the <dfn>set of filter properties</dfn> be the <a href="http://www.w3.org/TR/WebIDL/#dfn-dictionary-member">dictionary members</a> of `entryProps`, or the empty set if `entryProps` is omitted.</li>
+              <li>For each available <a>PerformanceEntry</a> object (<dfn>entryObject</dfn>), in chronological order with respect to <a href="#widl-PerformanceEntry-startTime">startTime</a>:</li>
+                <ol>
+                <li>For each _key_ and _value_ pair in <a>set of filter properties</a>:</li>
+                  <ol>
+                  <li>If the <a>entryObject</a> does not contain an attribute whose name matches _key_, go to next <a>entryObject</a>.</li>
+                  <li>Otherwise, if the <a>entryObject</a> contains an attribute who's name matches _key_, and its value does not match _value_, go to next <a>entryObject</a>.</li>
```
lawnsea: You missed the second s/who's/whose/: `an attribute who's name matches`

lawnsea:
:+1: other than the remaining nit

toddreifsteck:
One of our developers, Justin Rogers, made a solid verbal argument against exposing these APIs in favor of default ES6 filter. Pinging him to comment on this thread.

DigiTec:
### Problem Statement: ###
It is hard to add new fields to the PerformanceData object since there is no easy way to get entries that match a given criterion. In the past specific filter methods were added to make this easier that could filter by type or name. As we add more fields this becomes less and less useful. In addition there is a performance concern about retrieving all entries and inducing garbage collection.

_Issue #1: Can only filter by type and by name given the current set of APIs_

_Issue #2: Returning all entries creates a lot of temporary objects that then have to be GC’ed post filtering. __We believe this issue is irrelevant given modern GC behavior.___

### Proposed Solutions: ###
#### Solution #1 (non-ideal): ####
Provide getEntries with a dictionary. Enumerate the keys of the dictionary and filter the data based on equality matching between key to performance data field name and field value to performance data value.

__Pros:__ Proposed solution easily allows the getEntries method to scale as new fields are added. 

__Cons:__ Proposed solution misuses the dictionary type. The dictionary type is designed to strongly type through duck typing, the incoming object against the expected results. A dictionary generally specifies names of supported fields and their types. It also specifies defaults in some cases. By not specifying these and enumerating all keys we use the dictionary in a way that is complicated for browsers to get right and standardize their behavior on.

Enumeration of keys in the dictionary is questionable. The dictionary is accessed using the [[Get]] method. Which means it obeys prototype inheritance. And you can even imagine that there would be some inheritance in filters. This means the enumeration has to be for all properties and not just own properties of the object. If the user passes in a large object, such as window or an element, then the browser will spend a large amount of time enumerating and even throwing exceptions processing all of the key/value pairs.

#### Solution #2 (non-ideal): ####
Provide a callback function to process each entry. This way the user can filter using complicated filter conditions such as ranges, complex conversions, regular expressions, etc…

__Pros:__ Allows the user to filter however they want.

__Cons:__ Proposed solution has a performance issue in that we have to make callouts during enumeration. Further, we still have to construct an object per entry which will then have the same GC characteristics of returning all entries.
We hand out objects during enumeration which means the author can hold onto them. They can also attach more data to them during the callback. Not sure if this is a pro or con.

#### Solution #3 (recommended): ####
Provide a subset of values as a dictionary, such as name and type, which are both typed to DOMString so the browser can do proper conversion. Use this as an initial filter. On the returned entries allow the user to use the javascript filter method to do further reduction. 

__Pros:__ Solves initial problem of being able to filter based on the set of supported properties.
Additional supported dictionary values over time to improve native filtering via the dictionary.

__Cons:__  Proposed solution in this case will fail whenever a user expects filtering for new members which the given browser doesn’t support. This may or may not be an issue as the user can still do additional post filtering.

toddreifsteck:
**Solution 3 summary:**
dictionary PerformanceEntryFilter {
  DOMString name = undefined;
  DOMString type = undefined;
};
getEntries(optional PerformanceEntryFilter filter);

The initial names are "name" and "type".

We can then add new names to PerformanceEntryFilter as we standardize various new attributes.

igrigorik:
@DigiTec @toddreifsteck thanks, this is great feedback! A couple of followup questions...

> We can then add new names to PerformanceEntryFilter as we standardize various new attributes.

Why would that be necessary? My understanding is that the application can specify any value for "name" and "type". Is that not the case?

Would it make sense to consider allowing a sequence of PerformanceEntryFilters, such that you can filter on multiple attributes?

DigiTec:
@igrigorik Ah, the original filter dictionary used names that are too generic. I apologize for that, it is very confusing. The proposal, isn't that name = "name of field" and type = "value of field" or anything like that. It is that if "name" is specified in the dictionary, that you filter the name property of the internal data objects by the given value. The same for type, if specified, you would filter the type property of the internal data objects. So each line in the dictionary for PerformanceEntryFilter represents a field in the internal data object that the browser can filter on. 

Note, that we wouldn't be able to filter on arbitrary values this way, only those we know about and only those we choose work well with direct equality comparisons (or if we so choose we can specify indirect fields in the dictionary such as durationStartRange/durationEndRange and that could do a range based filter of anything in the duration length).

For more complicated filters, the user would still do a .filter() on the returned PerformanceEntryList. I would recommend not using this typedef, since it makes it unclear that the return type is actually a sequence<> -> which is a JavaScript Array, as opposed to a Platform Object. The list nomenclature makes me think that this "list" might have similar semantics to say an HTMLCollection (live, lazy, etc...) which it doesn't. It might also make me think I can't use array methods directly and would then have to favor using the ugly long form Array.prototype.filter.call(getEntries(), function() {}) which is the duck typed method for dealing with Platform Object collections that have a length parameter.

I guess my intent was to supply a complicated enough filter (allowing them to filter on any known, supported parameters, and as many as they choose) to remove the necessity for things like getEntriesByType and getEntriesByName, while at the same time ensuring that they can't filter on arbitrary name/value pairs since that is difficult for the browser to implement as a dictionary type without potential performance hits.

If, we decided that we really want to support arbitrary name value pairs, then I would counter propose the following:

dictionary PerformanceEntryFilter {
    sequence<DOMString> keys;
    sequence<DOMString> values;
}

The user would then just specify an object which contained an array of key names (field names) and an array of values that the fields have to match. This doesn't in any way solve the ranges or complex comparisons but they can still do that as a post filter.

So options:
1. Fixed function filter. You can filter by any name that is in the browser's supported dictionary. Additional filtering of arbitrary fields not in the filter dictionary would have to be done post retrieval.
2. Provide sequences of keys (or fieldNames) and values in a dictionary type instead. This allows for arbitrary filtering, but will have odd side effects like me supplying a fieldName that doesn't exist and therefore all entries get filtered out. Because of the arbitrary nature of the fieldNames we run the risk of browser interop issues if someone writes a filter that only works on one browser and not another do to the addition of a bleeding edge field name.

igrigorik:
Ah, I see, I think misread Todd's example. Just to confirm that we're on the same page... 

The original proposal is that we define a dictionary with `name` and `entryType` keys, and allow the application to specify the corresponding (DomString) values. Concretely: `getEntries({name: x, entryType: y, bogus: z})` -- the `bogus` field would be ignored because it's not a supported key; you can combine multiple supported keys to create more complex filters. Is that right? In effect, we're simply whitelisting a list of "known/supported" keys that can be filtered on?

DigiTec:
@igrigorik Yes, that is exactly right! As time goes on we support and white-list more things that the browser might support filtering on. In the short term, poly-fills could do the additional filtering while we are prototyping new properties that aren't broadly adopted yet.

igrigorik:
@DigiTec hmm, are there any other ways to mitigate your concerns, short of a whitelist? E.g. if we allow User Timing events to have custom payloads/attributes, I'd be nice to allow efficient filtering on them: https://github.com/w3c/user-timing/issues/3. We should probably take a closer look at that thread before we resolve this one.

DigiTec:
@igrigorik The assumption is that the browser can filter faster than the JavaScript. I don't think that is true. While we can implement some basic filtering for known properties extremely efficiently (internally we can use hash-tables for instance to group by name/entryType), once you generalize the problem most of our efficiency will be lost. Also, we assume that the only comparator for filtering is equality, which I also argue is not the only case. I'm likely to have complex attribute/payloads that need some form of sub-parsing. Whether the comparators are begins with, ends with, case insensitive, regex, range, etc... I don't think equality will be the core use case.

I agree that having filtering of user defined values is a valuable thing. I just think that we are better off providing the user with the building blocks that allows them to build out all of their filtering needs. The day that we defined this as sequence<> we did just that. We ensured that every filtering primitive on the planet could work. jQuery, lo-dash, you name it. They can all be used and will be exceedingly fast at reducing the list even with complex comparators.

natduca:
The browser can filter faster than javascript. The key thing is that to send an event into javascript, we must construct the javascript object. If you tell us that you're not interested in a particular event signature, we can simply skip creating the object [and even skip buffering it when its event happens]. This property of "discard before JS" is critical for good performance.

DigiTec:
@natduca This is a pull method. I think you might be talking about performance observers, which would allow you to state which objects you wanted and then browsers can choose to buffer or not. For getEntries, we have to process all buffered events relative to your filter. So there is no concept of not buffering them and still supporting this API unless I'm mistaken in some way about your point. Which I could be so feel free to correct me.

Modern GCs can allocate/collect a LOT of objects per second. It is very unlikely that the microseconds it takes us to cook these up will affect your performance. You need to be in the hundreds of thousands or querying many times per second, all of which seem unnecessary.

igrigorik:
@DigiTec fwiw, we do have the concept of [LazyPerfEntryList](http://w3c.github.io/performance-timeline/#the-lazyperformanceentrylist-interface) that we've added in PerfObserver, the intent behind which is to allow us to emit the list without necessarily materializing the underlying entry objects in JS-space. To get the actual entries you'd have to call `getEntries`, at which point the query syntax becomes relevant.. as it determines what/how much we need to move from browser side into JS space. That said, I take your point about perf of modern GC's.. 

I'll defer to @natduca on overhead -- he's been the closest to our tracing implementation.

DigiTec:
Per the conference call, we wanted to formalize the ideas for a) having a fixed dictionary for filtering and comparison and b) having the option for arbitrary filter values for use by user timings if we add a generic user attributes later.

Here is our default filter where we'd specify all of the existing types. During the conference call someone mentioned about DOMHighRestTimestamp matching. I noted that conversion to string and comparing as string would work, including for the various short fields. However, DOMString is only required for custom values, not for our own defined values, so we can strongly type them if we wish and internally do strongly typed checks as well.

```
dictionary PerformanceEntryFilter {
    DOMString name;
    DOMString entryType;
    DOMHighResTimestamp startTime; // We can support strongly typed matching on our own types
    // Continue enumerating all valid existing fields for the performance data types
}
```

Here is an extension to the filter which allows the specification of custom keys and values. There is no way to further restrict these using WebIDL so we should add prose about how these are used to be used and under what circumstances the browser will either throw or ignore. Not sure what that prose would look like yet, but we can work on it.

```
partial dictionary PerformanceEntryFilter {
    // Extension points for having user specified name/value pairs be filtered
    // Prose recommendation that the length of customKeys and customValues be the same.
    sequence<DOMString> customKeys;
    sequence<DOMString> customValues;
}
```

igrigorik:
@DigiTec thanks for the summary.

Reviewing the current list of [PerformanceEntry metrics](http://www.w3.org/wiki/Web_Performance/EntryType), I'm wondering if the union of all attributes is the route we want to go down. For example, take `startTime`, or any one of the other HighRes timestamps: while we can compare on them, I can't think of a single use case where you'd want to filter on them with an equality operator at microsecond granularity.. you'd end up with just that one entry, or random sample of entries. Practically speaking, for timestamps you probably want a range operator, and I think that's out of scope for this interface (and can be done in client-space).

As such, the minimum viable solution (to match what we have today) here is to define:

```
dictionary PerformanceEntryFilter {
    DOMString name;
    DOMString entryType;
}
```

Based on the above, we can explain current getEntriesBy{Name,Type} methods in the same way as noted in https://github.com/w3c/performance-timeline/pull/11#issue-72314087. From there, we can then start extending the whitelist as necessary to match new use cases as they come up. For example.. add [sourceFrameNumber](http://w3c.github.io/frame-timing/#widl-PerformanceRenderTiming-sourceFrameNumber)?

DigiTec:
@igrigorik Defining the replacement for the existing expanded methods seems like the right choice. I'm liking this a lot!

igrigorik:
@DigiTec ptal: https://github.com/w3c/performance-timeline/commit/30d649f50020adb7a107bb481bd90eaa6c06c9db

Preview: https://rawgit.com/w3c/performance-timeline/getEntries-filter/index.html#the-performance-interface (see EntryProps and getEntries methods).

igrigorik:
@plehegar @toddreifsteck I believe this should be good to land now. PTAL, let me know if you see any issues.

igrigorik:
Discussed on the WG call on 6/24/15: merging, will ping the mailing list for further feedback.

PR #12, opened 2015-05-11
(WIP) Observer-based processing

igrigorik:
Performance Timeline does not specify any processing steps, or provide hooks to other interfaces, to register and emit new metrics. This pull seeks to address both of these. Preview:

https://rawgit.com/w3c/performance-timeline/observer-processing/index.html#performance-timeline

After a few attempts at different language, I'm proposing that we define processing in terms of interaction with the `PerformanceObserver` interface: register -> (optional) buffer -> notify. The old `Performance` interface can be polyfilled directly on top of `PerformanceObserver` via JavaScript, either by the UA or the application - e.g. subscribe to each supported metric type, apply relevant buffering strategy, etc.

Thoughts, comments?

### DIFF
--- a/index.html
+++ b/index.html
@@ -123,21 +123,40 @@
   </section>
 
   <section>
-    <h2><dfn>Performance Timeline</dfn></h2>
+    <h2>Performance Timeline</h2>
 
-    <p>The <a>Performance Timeline</a> enables the user agent and application developers to access, instrument, and retrieve various performance metrics from the full lifecycle of a web application.</p>
+    <p>Each browsing and worker context has an associated <dfn>performance timeline</dfn>. The <a>performance timeline</a> is responsible for processing of registered metrics, providing methods to retrieve metrics via the <a>Performance</a> interface, and delivery of notificiations to relevant subscribers via the <a>Performance Observer</a> interface.</p>
 
-    <p>All interfaces that participate in the <a>Performance Timeline</a> MUST adhere to the following rules:</p>
+    <section>
+      <h2>Processing</h2>
 
-    <ul>
-      <li>MUST extend the <a>PerformanceEntry</a> interface</li>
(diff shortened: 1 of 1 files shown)


igrigorik:
Merging feedback from https://github.com/w3c/performance-timeline/issues/14 (cc @esprehn)

> The spec needs to specify a task source, and how the callbacks are queued, and when they're called. It also needs to specify what the callback this value is. Please see MutationObservers for how to correctly spec this:
> https://dom.spec.whatwg.org/#mutation-observers
> https://heycam.github.io/webidl/#dfn-callback-this-value
> 
> The spec needs detailed steps for what to do in each step and how to execute each algorithm.
> 
> Note that specifically it seems like from one of the examples that you might get synchronously notified of existing records which seems bad. The spec should talk about how and when to queue tasks for notify you.

igrigorik:
Every observer has an internal buffer, and for every event we insert into appropriate buffers. That will resolve https://github.com/w3c/performance-timeline/issues/16 and remove ambiguity of existing records.

index.html
```diff
@@ -123,21 +123,40 @@
...
+          <ul>
+          <li>The user agent SHOULD seek to minimize overhead and resource contention with other and more critical application logic. For example, buffer one or more registered metrics prior to notifying the <a>registered performance observer</a>'s.</li>
+          <li>The user agent MAY define own buffering logic. For example, it may periodically process the buffered metrics, based on a timer, number of metrics in the buffer, and other criteria.</li>
+          </ul>
+
+        <li>To <dfn>notify an observer</dfn>, the user agent MUST execute the following steps:</li>
+          <ol>
+          <li>For each <a>registered performance observer</a> (_observer_):</li>
+          <ol>
+            <li>Let the <dfn>lazy list of entry objects</dfn> be the empty <a>LazyPerformanceEntryList</a>.</li>
```
mpb: We removed the 'Lazy' prefix.

igrigorik:
Closing. Superseded by https://github.com/w3c/performance-timeline/pull/46.

Issue #15, opened 2015-05-30
Second example disconnects immediately which means nothing is delivered


esprehn:
ex. https://w3c.github.io/performance-timeline/#h-introduction

has:

```js
var observedEvents = [];
var observer = new PerformanceObserver(function(list) {
  observedEvents.push(list); // defer processing
});

// subscribe to Frame-Timing and User-Timing events
observer.observe({eventTypes: ['render', 'composite', 'mark', 'measure']});
observer.disconnect();
}
```

which disconnects immediately after observing which means you should get nothing delivered.

mpb:
Better?

PR #21, opened 2015-06-05
Fix: #15 Second example disconnects immediately

mpb:
Fix: #15 Second example disconnects immediately which means nothing is delivered

### DIFF
--- a/index.html
+++ b/index.html
@@ -103,27 +103,35 @@
     <p>Alternatively, instead of processing metrics at a predefined time, or having to periodically poll the timeline for new metrics, the developer may also observe the <a>Performance Timeline</a> and be notified of new performance metrics via a <a>Performance Observer</a>:</p>
 
     <pre class="highlight example">
-    var doneObservingEvents = false;
-    var observedEvents = [];
+&lt;!doctype html&gt;
+&lt;html&gt;
+  &lt;head&gt;
+  &lt;/head&gt;
+  &lt;body onload="init()"&gt;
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -110,7 +111,6 @@
 
     // subscribe to Frame-Timing and User-Timing events
     observer.observe({eventTypes: ['render', 'composite', 'mark', 'measure']});
-    observer.disconnect();
 
     // processing logic called by application code
     function processObservedPerfEvents() {
```
igrigorik: The example never calls `processObservedPerfEvents()`.. The idea was to provide a "defer processing" example, but FWIW, perhaps that's unnecessary as a first intro to the API. My proposal would be to merge contents of that function into the PO callback?

mpb: PTAL

index.html
```diff
@@ -103,23 +103,35 @@
     <p>Alternatively, instead of processing metrics at a predefined time, or having to periodically poll the timeline for new metrics, the developer may also observe the <a>Performance Timeline</a> and be notified of new performance metrics via a <a>Performance Observer</a>:</p>
 
     <pre class="highlight example">
-    var observedEvents = [];
+&lt;!doctype html&gt;
+&lt;html&gt;
+  &lt;head&gt;
+  &lt;/head&gt;
+  &lt;body onload="init()"&gt;
```
igrigorik: onload is unnecessary, otherwise lgtm!

mpb: Better?

igrigorik:
lgtm, thanks!

Issue #40, opened 2015-07-20
window.performance.createObserver feedback


igrigorik:
https://github.com/w3c/performance-timeline/pull/29 landed `window.performance.createPerformanceObserver`, but we have a few -1's on the new API:

> PerformanceObserver should lookup the |performance| object in the context on which it's defined. That's how new Text() gets an ownerDocument, how new EventSource() finds an origin, how new Notification() works, how new Worker() gets an origin... most of the web platform is based around the idea of looking up objects in the current context -- https://github.com/w3c/performance-timeline/pull/29#issuecomment-122602246.

> Strong -1 for changing from new PerformanceObserver to window.performance.createObserver -- https://github.com/w3c/performance-timeline/pull/29#issuecomment-122706223 

---

To lookup the performance object we can reference the "[entry settings object](https://html.spec.whatwg.org/multipage/webappapis.html#entry-settings-object)" ([EventSource example](https://html.spec.whatwg.org/multipage/comms.html#the-eventsource-interface)).

Any other implementation or other concerns with this API? Should we go back to `new PerformanceObserver`?

/cc @bzbarsky @esprehn @domenic

bzbarsky:
Using the entry settings object would be _really_ weird.  That's not what `new Text` does, for sure.  It can sometimes make sense for base URIs, less often for origins, never for anything else.

If you want to do what `new Text` does, I suspect that's using the global of the Realm of the `Text` function.  At least in browsers.  The spec for `new Text` doesn't really define the behavior because it talks about "the global object", which is not a defined concept in general.  /cc @annevk

I still think the implicit magic is bad, but could live with it as long as it's defined properly.

annevk:
We really need better hooks from HTML and IDL. Including some guidelines. Or a pattern that people can copy.

bzbarsky:
Yes, we do.

igrigorik:
@annevk @bzbarsky I'd appreciate any suggestions for how to spec this. Do we have any existing open issues to define such hooks?

annevk:
I filed bugs against IDL and also HTML, but W3C Bugzilla is down at the moment.

igrigorik:
Great, thanks! It'll be great to have URLs to those. 

In the meantime, since folks are already implementing this interface, should we revert back to `new PerformanceObserver` and add a note to clarify the spec language once above bugs are resolved?

annevk:
Why are we doing this API when we'll get `FetchObserver`?

[...]

PR #43, opened 2015-07-24
Revert to using a constructor for PerformanceObserver.

mpb:
As per #40, revert from window.performance.createPerformanceObserver to using new PerformanceObserver

### DIFF
--- a/index.html
+++ b/index.html
@@ -128,7 +128,7 @@
     });
 
     // subscribe to Frame-Timing and User-Timing events
-    observer.observe({entryType: ['render', 'composite', 'mark', 'measure']});
+    observer.observe({entryTypes: ['render', 'composite', 'mark', 'measure']});
     &lt;/script&gt;
   &lt;/body&gt;
 &lt;/html&gt;
@@ -233,7 +233,7 @@
         <dt>void observe(PerformanceObserverInit options)</dt>
         <dd>This method instructs the user agent to <dfn>register the observer</dfn> and report any new performance entries based on the criteria given by <a>options</a>.
           <dl title='dictionary PerformanceObserverInit' class='idl'>
-            <dt>required sequence&lt;DOMString&gt; entryType</dt>
+            <dt>required sequence&lt;DOMString&gt; entryTypes</dt>
             <dd>
               <p>A list of valid <a href="#widl-PerformanceEntry-entryType">entryType names</a> to be observed. The list MUST NOT be empty and types not recognized by the user agent MUST be ignored.</p>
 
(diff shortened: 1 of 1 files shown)


igrigorik:
Hmm, we'll have a conflict with https://github.com/w3c/performance-timeline/pull/37 -- that pull changes typeFilter to entryType to be consistent with getEntries*. But that's more of a merge concern.. 

Perhaps we should add a note indicating that the language should be updated once https://github.com/w3c/performance-timeline/issues/40#issuecomment-122975250 is resolved? 

/cc @bzbarsky @annevk

igrigorik:
> I filed bugs against IDL and also HTML, but W3C Bugzilla is down at the moment.

@annevk do you have links to these?

annevk:
See https://www.w3.org/Bugs/Public/show_bug.cgi?id=24652 and its dependencies.

mpb:
Resolved the conflict with #37 
Although `entryType` sounds a little funny for a sequence name.
```WebIDL
    required sequence<DOMString> entryType;
};
```

index.html
```diff
@@ -225,42 +225,35 @@
...
+            </dd>
+          </dl>
+
+          <p>The `observe(options)` method must run these steps:</p>
+          <ol>
+            <li>If _options'_ `entryType` attribute is not present, <a href="http://heycam.github.io/webidl/#dfn-throw">throw</a> a JavaScript `TypeError`.</li>
+            <li>Filter unsupported <a href="#widl-PerformanceEntry-entryType">entryType names</a> within the `entryType` sequence, and replace the `entryType` sequence with the new filtered sequence.</li>
+            <li>If the _options'_ `entryType` attribute is an empty sequence, <a href="http://heycam.github.io/webidl/#dfn-throw">throw</a> a JavaScript `TypeError`.</li>
+            <li>Add a new <a>registered performance observer</a> with the <a href="http://www.w3.org/TR/dom/#context-object">context object</a> as the <dfn>observer</dfn> and _options_ as the `options`.</li>
```
igrigorik: I'm a bit confused by this last line.. "context object as the observer"?

toddreifsteck:
Per the 8/12 call, altering entryType to entryTypes likely makes sense given it is more than one.

mpb:
PTAL

igrigorik:
Looks good, thanks Michael!

Based on the discussion on the WG call today, I think we're all in agreement that the `new PerformanceObserver` interface is what we want.. And there has already been some confusion around this as current spec is showing the wrong pattern. With that in mind, I'm merging this to get the right interface in front of folks reading the spec, and I've opened a separate issue ( see https://github.com/w3c/performance-timeline/issues/44) to clarify the spec language.

PR #46, opened 2015-08-19
Processing algorithm for PerformanceObserver

igrigorik:
Preview: https://rawgit.com/w3c/performance-timeline/processing/index.html#performance-timeline

- Provides a "queue a PerformanceEntry" hook for other specifications.
- Defines steps for delivering events to observers.
- Defines task to trigger application callbacks for observers with buffered entries.

The rough algorithm is as follows: events are queued by various subsystems and delivered into each observer's buffer; when an event is queued, a task is created to trigger application callback on relevant observers; the task iterates over all registered observers and dispatches application callbacks for observers that have buffered events.

### DIFF
--- a/index.html
+++ b/index.html
@@ -148,7 +148,7 @@
   <section>
     <h2><dfn>Performance Timeline</dfn></h2>
     <p>Each <a href=
-    "https://html.spec.whatwg.org/multipage/browsers.html#unit-of-related-similar_origin-browsing-contexts">
+    "https://html.spec.whatwg.org/multipage/browsers.html#unit-of-related-similar-origin-browsing-contexts">
     unit of related similar-origin browsing contexts</a> has a <dfn>performance
     observer task queued flag</dfn> and an associated list of <a>registered
     performance observer</a> objects which is initially empty.</p>
@@ -190,7 +190,7 @@
           <li>Unset <a>performance observer task queued flag</a>.
           </li>
           <li>Let <i>notify list</i> be a copy of <a href=
-          "https://html.spec.whatwg.org/multipage/browsers.html#unit-of-related-similar_origin-browsing-contexts">
+          "https://html.spec.whatwg.org/multipage/browsers.html#unit-of-related-similar-origin-browsing-contexts">
             unit of related similar-origin browsing contexts’</a> list of
             <a>registered performance observer</a> objects.
           </li>
--- a/index.html
+++ b/index.html
@@ -147,31 +147,80 @@
   </section>
   <section>
     <h2><dfn>Performance Timeline</dfn></h2>
-    <p>The <a>Performance Timeline</a> enables the user agent and application
-    developers to access, instrument, and retrieve various performance metrics
-    from the full lifecycle of a web application.</p>
-    <p>All interfaces that participate in the <a>Performance Timeline</a> MUST
-    adhere to the following rules:</p>
-    <ul>
-      <li>MUST extend the <a>PerformanceEntry</a> interface
+    <p>Each <a href=
+    "https://html.spec.whatwg.org/multipage/browsers.html#unit-of-related-similar_origin-browsing-contexts">
+    unit of related similar-origin browsing contexts</a> has a <dfn>performance
+    observer task queued flag</dfn> and an associated list of <a>registered
+    performance observer</a> objects which is initially empty.</p>
+    <p>To <dfn>queue a PerformanceEntry</dfn> (<i>new entry</i>), run these
+    steps:</p>
+    <ol>
+      <li>Let <i>interested observers</i> be an initially empty set of
+      <a>PerformanceObserver</a> objects.
       </li>
-      <li>MUST be supported by the <a href=
-      "#widl-PerformanceObserverEntryList-getEntries-PerformanceEntryList-PerformanceEntryFilterOptions-filter">
-        getEntries</a>, <a href=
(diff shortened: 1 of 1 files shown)


hiikezoe:
> If observer's PerformanceObserverInit entryTypes includes new entry’s entryType value, append observer to interested observers. 

I think this filtering process should be done in the queued task.
 Otherwise, the observer in below code receives unexpected entries.

observer.observe({entryTypes: ['mark']});
performance.mark('test');
observer.observe({entryTypes: ['measure]}); // after this, observer should received only PerformanceMeasureEntry.

hiikezoe:
I was wrong.

> observer.observe({entryTypes: ['mark']});
> performance.mark('test');

This |mark| entry should be notified to the observer. Because the |mark| event has been done before observe method is called with a new options.
I am sorry for the confusion.

igrigorik:
@hiikezoe that would cause some interesting and -- I think -- unwanted repercussions:

1. It would mean that every observer would have to buffer every event, regardless of its currently specified filter options.
1. It would mean that every interface participating in the Perf Timeline would have to always emit events, regardless of whether there are any active observers looking for those events or not.

We could probably address (1), but we can't do anything about (2), and I think that's a deal breaker. We should operate on the assumption that we'll have _many_ interfaces participating in the timeline, and that these interfaces may emit _large numbers_ of events.. As such, we want to reserve the right to omit queueing such events at all if no observers are looking for them.

> This |mark| entry should be notified to the observer. Because the |mark| event has been done before observe method is called with a new options.

Yep, and that's what current draft will deliver.

hiikezoe:
I noticed another edge case:

 observer.observe({entryTypes: ['mark']});
 performance.mark('test');
 observer.disconnect();

The mark entry should be notified to users. An easiest way to achieve this is to call the callback in disconnect() process if observer buffer is not empty, but I don't think it is good idea. We should have another observer list having observers which have non-empty observer buffer even if the observer has been already disconnected.

igrigorik:
> We should have another observer list having observers which have non-empty observer buffer even if the observer has been already disconnected.

As in, to allow the observer get the callback with buffered events after calling disconnect()? That seems a bit weird.

FWIW, I think MutationObservers address this via [takeRecords](https://dom.spec.whatwg.org/#dom-mutationobserver-takerecords), which allows the application to empty the buffer before calling disconnect. That said, I think @natduca had some reservations about exposing that for PerformanceObserver... Nat, any thoughts?

natduca:
In the case of mark events, its maybe plausible. An interesting one to think through though is something where the actual event doesn't happen synchronously. E.g. we begin a redraw of the screen, but then disconnect. Do we get frame events for that in-progress frame?

It seems like either we go a route of saying which perf events are delivered synchronously vs asynchronously, or we say that they're all delivered async. Its a hard space.

igrigorik:
Hmm, so from a layering perspective... As far as Performance Timeline is concerned, there is no such thing as "sync vs async events": the timeline only provides a method to queue an event, it has no say on when participating interfaces queue the events. Once the event is queued it goes into appropriate observer buffers, and a task is queue to notify observers.

What @hiikezoe highlighted in https://github.com/w3c/performance-timeline/pull/46#issuecomment-132888912 is that it's possible to call `disconnect()` in the interval between the time when the event was queued and when the notification task fires... Which means there might be events in the observer buffer that were registered but that would be cleared without delivery.

That said, now that I think about it, I think that's easy to mitigate: call `disconnect()` from within the notification callback. @hiikezoe I think that's sufficient?

hiikezoe:
> I think that's easy to mitigate: call disconnect() from within the notification callback. @hiikezoe I think that's sufficient?

You mean disconnect() is restricted in the notification callback? It sounds inconvenient. I am sorry if I don't understand what you mean.

igrigorik:
No, there is no reason to restrict it. Rather, if you -- that is, the application developer -- really care about making sure that the observer buffer is clear before disconnecting, you can wait for the next callback and call disconnect from within it.

hiikezoe:
Hmm... I am still guessing it's inconvenient to users even if it's not restriction.
@eliperelman I would like to know what you think because I think you have been using performance APIs heavily.

esprehn:
We shouldn't ever call the callback after a ```disconnect()```, we can add a takeRecords() if you want for that?

Calling ```disconnect()``` means you're not interested in notifications anymore, don't do that if you still care about what's in the buffer.

eliperelman:
Thinking about the normal flow of these types of things from a web developer perspective, I would expect that calling `disconnect()` should not trigger any more observations. If we want to ensure those marks are delivered, then disconnect should not be in a race with mark.

So in this case, I would agree with @igrigorik and @esprehn.

eliperelman:
And by "not being in a race", I mean the userland code should be written to mitigate that.

igrigorik:
@mpb @eliperelman @esprehn @hiikezoe @toddreifsteck any objections to merging this? With this in place we finally have the hooks to start updating other specs.

If we want to pursue takeRecords, we can tackle that as a separate discussion (so far, I'm not convinced it's strictly necessary).

toddreifsteck:
"unit of related similar-origin browsing contexts" seems like it could be better defined. When I reviewed HTML5 spec link, it did not seem to have the concept of "related similar-origin" defined and only referred to browsing contexts. Am I missing something?

Everything else seems fine to me.

igrigorik:
@toddreifsteck oops, it looks like I broke the link for that. Fixed in f2caba9, ptal.

mpb:
This looks ok to me.

toddreifsteck:
LGTM! Merging

PR #50, opened 2015-11-02
Observers observe user timing entries.

tdresser:
The previous phrasing made it sound as though only performance timeline entries which the user agent is responsible for recording are observable. It might be better to make this more explicit than what I've proposed, but I think this is less confusing than what was there before.


igrigorik:
Hmm, what's the issue with the "by the user agent"? I don't see any issues with wording it either way, but want to make sure I'm not missing anything here.

tdresser:
I initially thought this meant that performance timeline entries added via performance.mark and performance.measure wouldn't be observed.

We should make it clear that they are.

igrigorik:
Ah, I see. Fair enough, merging.

PR #52, opened 2016-04-12
Explain why PerfObserver in the introduction

igrigorik:
Adds some additional context for why PerfObserver interface was added,
what problems it addresses, and why it should be the preferred method to
access the Performance Timeline moving forward.

---

Context: [TAG feedback](https://github.com/w3ctag/spec-reviews/issues/18#issuecomment-206523833)

/cc @toddreifsteck @plehegar @yoavweiss

### DIFF
--- a/index.html
+++ b/index.html
@@ -129,10 +129,9 @@
   &lt;/body&gt;
 &lt;/html&gt;
 </pre>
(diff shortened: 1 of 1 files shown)


plehegar:
:+1:

PR #70, opened 2017-02-24
add a note regarding feature detection

### LINKED ISSUE #64: Add a note regarding feature detection options for new entry types
While reviewing a LongTask implementation [patch](https://codereview.chromium.org/2610063006/) I noticed that there's no way for developers to detect browser support for new entry types:

* In [Filter performance entry buffer by name and type](https://w3c.github.io/performance-timeline/#filter-performance-entry-buffer-by-name-and-type), we just return an empty list for unknown types.
* In [The PerformanceObserver interface](https://w3c.github.io/performance-timeline/#the-performanceobserver-interface), we specify "types not recognized by the user agent must be ignored" and the processing model later filters and ignores unsupported types.

I think we should change that so that at least one but ideally both throw or return an error when called with an unsupported entry type.

/cc @spanicker @igrigorik

igrigorik:
closes #64.

@yoavweiss ptal.

### DIFF
--- a/index.html
+++ b/index.html
@@ -400,12 +400,6 @@ interface PerformanceObserver {
           entryType names</a> within the `entryTypes` sequence, and replace
           the `entryTypes` sequence with the new filtered sequence.
         </li>
-        <p class="note">In some cases the developer may want to distinguish
-        between absence and lack of support for a particular type of entry. In
-        such cases they can feature detect the presense of the relevant
-        interface—e.g. to detect if User Timing entries are supported, the
-        developer can test for `PerformanceMark` and `PerformanceMeasure`
-        interfaces.</p>
         <li>If the _options'_ <a for="PerformanceObserverInit">entryTypes</a>
         attribute is an empty sequence,
         [throw] a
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -400,6 +400,12 @@
           entryType names</a> within the `entryTypes` sequence, and replace
           the `entryTypes` sequence with the new filtered sequence.
         </li>
+        <p class="note">In some cases the developer may want to distinguish
```
yoavweiss: It might be better to add the note outside of the processing model. (so right after the `</ol>`)

index.html
```diff
@@ -400,6 +400,12 @@
...
+        between absence and lack of support for a particular type of entry. In
```
yoavweiss: That first phrase is not very clear and doesn't cover polyfills. Maybe "In some cases developers may want to distinguish between the absence of an entry in the timeline and lack of support for that entry, or may want to conditionally create alternative measurements for unsupported entry types."

index.html
```diff
@@ -400,6 +400,12 @@
...
+        such cases they can feature detect the presense of the relevant
```
yoavweiss: s/feature detect/feature-detect/
s/presense/presence/

index.html
```diff
@@ -400,6 +400,12 @@
...
+        interface—e.g. to detect if User Timing entries are supported, the
```
yoavweiss: IIUC we're testing on the existence of the constructor, not the interface.

igrigorik:
@yoavweiss thanks, updated.. ptal.

yoavweiss:
LGTM

PR #84, opened 2017-07-24
Fix auto-publish

marcoscaceres:
[Auto-publishing is currently failing](https://lists.w3.org/Archives/Public/public-tr-notifications/2017Jul/0046.html) because it thinks you are trying to republish the document as a Working Draft. You should continue to make non-normative changes to this spec, and publish them as CR. 

Otherwise, if you are making substantive changes to the spec (that would warrant dropping back to WD), you might want to disable auto publishing entirely.

### DIFF
--- a/W3CTRMANIFEST
+++ b/W3CTRMANIFEST
@@ -1,1 +1,1 @@
-index.html?specStatus=WD;shortName=performance-timeline-2;useExperimentalStyles=false respec
+index.html?specStatus=CR;shortName=performance-timeline-2;useExperimentalStyles=false respec


igrigorik:
LGTM. @siusin can you sanity check?

siusin:
I don't think the auto-pub system can handle a living CR document... /cc @deniak @plehegar 

We can either make the changes in the editor's draft, and add them to the /TR version when we move to Proposed REC, or publish a FPWD of performance-timeline-3 and maintain the changes in the L3 WDs.

deniak:
True, the system can only handle some WDs and Notes for now. Republishing a CR needs to go to webreq.

marcoscaceres:
Hmmm, can we get that fixed? As long as the process is followed, there is no reason to not allow CR to automatically publish.

deniak:
Yes, the goal is to add more types of documents such as CRs and FPWDs but it's a bit tricky as there are other systems involved (IPP, comm announcement, etc) for these. So it's in the todo list but we just need to find time to work on the implementations.

igrigorik:
@siusin @plehegar bump. Any guidance here? Can we land this? If not, should we close this?

siusin:
Not much progress on Level-2 because it's still in CR, but we can auto-pub the WDs of Level-3 once the group make the decision to publish a FPWD.

igrigorik:
Since we're still actively iterating on L2.. Closing this, we can revisit once we have L3 FPWD.

Issue #87, opened 2017-09-12
Should PerformanceObserver throw when entryTypes is empty?


igrigorik:
PerformanceObserver is specced to ignore unknown entry types, and to throw a TypeError if the supported list of entries is empty. The former behavior is to enable extensibility, and the latter as a warning that you may be creating a no-op observer. However, there is a tension and inconsistency between the two, and we have two high profile cases where this caused web compat issues:

- https://github.com/webcompat/web-bugs/issues/7491 - Spotify wanted to observe "paint" but does not guard for empty list / raised exception.
- https://github.com/webcompat/web-bugs/issues/9716#issuecomment-328694263 - Google Cloud Console made the same mistake with "longtask". 

I suspect that we should [revisit the logic](url)...

> The observe() method instructs the user agent to register the observer and must run these steps:
> 1. Filter unsupported names within the entryTypes sequence, and replace the entryTypes sequence with the new filtered sequence.
> 2. If the options' entryTypes attribute is an empty sequence, throw a JavaScript TypeError.

As mentioned above, the intent behind step (2) is to warn the developer that they may be registering a zombie observer if none of the types are supported. We do have a separate request to enumerate supported task types (https://github.com/w3c/performance-timeline/issues/77), but I don't think that alone is sufficient to address the issue above — it's not obvious that I should manually check the types before requesting them, especially if unknown ones are ignored, and the gotcha case is when all of them are ignored.

As one option, should we consider dropping step 2? Is there a different way to signal that observer won't be receiving any entries and do we care? From an operational standpoint, it's a no-op object.

igrigorik:
@toddreifsteck @spanicker @rniwa @hiikezoe @JosephPecoraro would love to hear your thoughts and feedback on this one.

wisniewskit:
Perhaps it would be best for the call to `.observe` to return an array of the entryTypes the UA was able to subscribe, so that sites can opt into that knowledge or silently ignore it (as they seem to wish to do so right now)?

cvazac:
I think we should drop step 2 - forcing people to try/catch to feature detect doesn't feel right. 

Returning the filtered list per @wisniewskit would work. Question - if the returned list was empty, would developers still need to `.disconnect()`?

One reason to do this would be to polyfill. Here are my attempts:
[interface check](https://gist.github.com/cvazac/22194645f411a68a50c39c88f05faf2d)
[...]

Issue #77, opened 2017-05-23
Add entryTypes method to enumerate type of entries observable via PerformanceObserver


spanicker:
Currently developers have to "just know" the names of entryTypes for observe:
...observer.observe({entryTypes: ["foobar"]});

Would be nice if there was a method to enumerate them.
(Maybe there is a way and I just don't know)

igrigorik:
> Would be nice if there was a method to enumerate them.
> (Maybe there is a way and I just don't know)

There isn't. However, one can feature detect support pretty easily.. see: https://github.com/w3c/performance-timeline/issues/64

spanicker:
Right this is not about feature detection but rather about discoverability of what types exist.
So it's a smaller use-case - that's about convenience (although current types can be guessed)

igrigorik:
Yep. So far we've said that it's already possible to script on your own and hence the utility of exposing a native API for it is ~low.. Happy to be convinced otherwise. /cc @toddreifsteck @plehegar

igrigorik:
A quick proof of concept to test support for various perf interfaces:

```js
perfTypes = {
  "PerformanceResourceTiming": "resource",      // Resource Timing
  "PerformanceNavigationTiming": "navigation",  // Navigation Timing 
  "PerformanceMark": "mark",                    // User Timing
  "PerformanceMeasure": "measure",              // User Timing
  "PerformanceLongTaskTiming": "longtask",      // Long Tasks
  "PerformancePaintTiming": "paint",            // Paint Timing
  "PerformanceServerTiming": "server",          // Server Timing
}

var supportedPerfInterfaces = Object.keys(perfTypes)
  .filter(interface => typeof window[interface] === "function")
  .reduce((acc, key) => acc.concat(perfTypes[key]), []);

console.log(supportedPerfInterfaces)
```

_Note: one gotcha to the above is that availability of interface != it can be subscribed to by PerformanceObserver today. For that we'd probably need UA support._

jfsiii:
If N userland scripts all have this snippet, N must be updated when the list changes.

How likely is this list of interfaces to change? How would a developer know the list has changed?

igrigorik:
@jfsiii all good questions, but none of them are new as far as polyfills go.. yes interfaces can be updated, as long as you periodically pull in the snippet from some canonical source, you'd be fine. 

(That's not argument for not exposing this natively, but just pointing out that these are not new concerns)

hiikezoe:
[...]

PR #88, opened 2017-09-25
observing an empty sequence of entryTypes is a no-op

plehegar:
This is a proposed fix for #87

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/performance-timeline/perf-timeline/observe/typeerr.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/performance-timeline/ec04827...4e15343.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -329,7 +329,8 @@
         </li>
         <li>If the resulting '<a>entryTypes</a>' sequence is an empty sequence, abort
           these steps. The user agent SHOULD notify developers when the steps are
-          aborted. A console warning might be appropriate, for example.
+          aborted. A console warning listing ignored unsupported names in `entryTypes`
+          might be appropriate, for example.
         </li>
         <li>If the list of <a>registered performance observer</a> objects
         associated with the <a>ECMAScript global environment</a> of the
--- a/index.html
+++ b/index.html
@@ -327,8 +327,9 @@
         <li>Filter unsupported names within the `entryTypes` sequence, and
         replace the `entryTypes` sequence with the new filtered sequence.
         </li>
-        <li>If the resulting '<a>entryTypes</a>' sequence is an empty sequence, abort these steps.</li>
-        <p class='note'>User agents are encouraged to log this fact in a developer console, to aid debugging.</p>
+        <li>If the resulting '<a>entryTypes</a>' sequence is an empty sequence, abort
+          these steps. The user agent SHOULD notify developers when the steps are
+          aborted. A console warning might be appropriate, for example.
         </li>
         <li>If the list of <a>registered performance observer</a> objects
         associated with the <a>ECMAScript global environment</a> of the
--- a/index.html
+++ b/index.html
@@ -325,9 +325,10 @@
       <dfn>register the observer</dfn> and must run these steps:</p>
       <ol data-link-for="PerformanceObserverInit">
         <li>Filter unsupported names within the `entryTypes` sequence, and
-        replace the `entryTypes` sequence with the new filtered sequence.</li>
-        <li>If the _options'_ <a>entryTypes</a> attribute is an empty sequence,
-        <a>throw</a> a JavaScript `TypeError`.
+        replace the `entryTypes` sequence with the new filtered sequence.
+        </li>
+        <li>If the resulting '<a>entryTypes</a>' sequence is an empty sequence, abort these steps.</li>
+        <p class='note'>User agents are encouraged to log this fact in a developer console, to aid debugging.</p>
         </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -325,9 +325,10 @@
       <dfn>register the observer</dfn> and must run these steps:</p>
       <ol data-link-for="PerformanceObserverInit">
         <li>Filter unsupported names within the `entryTypes` sequence, and
-        replace the `entryTypes` sequence with the new filtered sequence.</li>
-        <li>If the _options'_ <a>entryTypes</a> attribute is an empty sequence,
-        <a>throw</a> a JavaScript `TypeError`.
+        replace the `entryTypes` sequence with the new filtered sequence.
+        </li>
+        <li>If the resulting '<a>entryTypes</a>' sequence is an empty sequence, abort these steps.</li>
```
igrigorik: Hmm, are we sure we want this here? As in, I think we still want the other steps to run?

igrigorik:
For reference, WPT update: https://github.com/w3c/web-platform-tests/pull/7474

index.html
plehegar: the observer will never get called since the list is empty. How would you differentiate from aborting the steps?

igrigorik: On a second and more careful read.. you're right — nevermind, all good.

index.html
```diff
@@ -325,9 +325,10 @@
...
+        <p class='note'>User agents are encouraged to log this fact in a developer console, to aid debugging.</p>
```
igrigorik: As a reference, looking at CSP, they actually make a normative "SHOULD" recommendation. 

> "The user agent SHOULD notify developers that a directive was ignored. A console warning might be appropriate, for example."

Perhaps we should do the same? Something like..

"... abort these steps. The user agent SHOULD notify developers that the observe call has aborted, due to empty '<a>entryTypes</a>' sequence. A console warning might be appropriate, for example."

plehegar: ignoring a CSP directive sounds more problematic noticing than failing to observe perf entries. it's probably worth asking around for opinions or not. I guess it depends on how strongly we feel. @yoavweiss , got an opinion as a dev?

yoavweiss: I think we should change the processing from filtering unsupported names to ignoring and logging them to console. A developer who is trying to observe "resource" and "navigation" and "longtaks" would want to know that "longtaks" entries are not coming anytime soon.

yoavweiss:
Would feature detection be done in a separate PR? If so, this is not closing #87...

index.html
plehegar: agreed, but how prescriptive should we be in logging into the console? My PR suggests a non-norm note, @igrigorik suggests a normative SHOULD.

plehegar:
@yoavweiss #87 is about throwing on empty list. it would be better to open a separate issue for feature detection imho.

igrigorik:
We already have https://github.com/w3c/performance-timeline/issues/77 to tackle feature detection — let's continue that conversation there.

> I think we should change the processing from filtering unsupported names to ignoring and logging them to console. A developer who is trying to observe "resource" and "navigation" and "longtaks" would want to know that "longtaks" entries are not coming anytime soon.

That's a great suggestion, it would help developers quickly identify unsupported types. Re, normative: it's a SHOULD clause, and creating empty observers can be misleading -- e.g. look, no long tasks in my code.. whereas, in reality the browser doesn't support logging those yet. My vote is for normative text, but with a SHOULD, logging unsupported types.

index.html
yoavweiss: My vote goes for a normative SHOULD.

plehegar:
ok. changed to use a normative should instead of the note.

index.html
```diff
@@ -325,9 +325,11 @@
...
       <ol data-link-for="PerformanceObserverInit">
         <li>Filter unsupported names within the `entryTypes` sequence, and
-        replace the `entryTypes` sequence with the new filtered sequence.</li>
-        <li>If the _options'_ <a>entryTypes</a> attribute is an empty sequence,
-        <a>throw</a> a JavaScript `TypeError`.
+        replace the `entryTypes` sequence with the new filtered sequence.
+        </li>
+        <li>If the resulting '<a>entryTypes</a>' sequence is an empty sequence, abort
+          these steps. The user agent SHOULD notify developers when the steps are
+          aborted. A console warning might be appropriate, for example.
```
igrigorik: How about: "A console warning listing ignored entryTypes might be appropriate, for example."

igrigorik:
Looks great, thanks Philippe! Merging.

Issue #81, opened 2017-06-19
Define buffering behaviors & hooks for queued (PerformanceObserver) entries


igrigorik:
We recently (see #76) added an optional flag to "queue an entry" that allows the caller to optionally append the entry to the [performance entry buffer](https://w3c.github.io/performance-timeline/#dfn-performance-entry-buffer).. For example, Long Tasks or Server Timing can set this to `true` while before onload to allow developers to register and retrieve any LT/ST records captured between start of page load and max(time when they register observer, onloadend). 

A few interactions that we did not address:
1. How does performance entry buffer interact with other buffers defined by individual specs? E.g. ResourceTiming defines own buffer with methods to query and clear it. 
  - When RT's buffer is cleared, does that affect the `performance entry buffer`?
1. Do we want to set a global cap on the `performance entry buffer`?
  - What happens when the limit is reached, do we start dropping items on the floor?

_Sidenote: moving forward we're not planning on adding more type-specific buffers.. Instead, we want to encourage use of PerfObserver. So, this is mostly an exercise in backwards compat and to make sure that we can pave a clean path for existing users of RT/NT/UT._

---

For (1), my hunch is "no", calls to clear type-specific buffers should not clear the performance entry buffer. This does mean that we might end up with some double book-keeping, but one of the motivating reasons for PerfObserver was to resolve the race conditions created by consumers clearing buffers from under each other. As such, I think I propose we treat them as separate entities: it should be possible to clear the RT buffer and still use PerfObserver with `buffered: true` to get all the entries queued before onload.

For (2), my hunch is "yes", and we should probably recommend a minimum ~ user agent should allow a minimum of XXXX entries to be queued, and once full the items are silently discarded.

/cc @cvazac @nicjansma @toddreifsteck

igrigorik:
Based on the [discussion at the WebPerf F2F](https://docs.google.com/document/d/1fcWpKZhJdrmBlpYUB2wIx42svo9zqojoJMZg8aY7E7Y/edit#bookmark=id.a3cjs2ikkil), we converged on: 

- Each spec defines a buffer with custom buffer size (depends on type of data, etc)
- Performance Timeline will buffer events until buffer is full. When the buffer is full...
  - Keep the first N entries, drop overflow
 - Buffer is **not** cleared at onload, but under memory pressure UA _may_ clear buffers
 - Some specs may provide additional methods to manipulate their buffer (e.g. ResourceTiming), but this is not a requirement.

[...]

Issue #74, opened 2017-03-17
Add support for flushing observer via `takeRecords()`


igrigorik:
Related discussion in https://github.com/w3c/performance-timeline/issues/66.

**Use case:** as a developer I want to flush and beacon all pending records when page visibility changes (e.g. transitions to 'hidden') or is about to be unloaded. Today, this is not possible.

To stay consistent with other `<*>Observer` patterns, we should support `takeRecords()`:
- [MutationObserver#takeRecords](https://dom.spec.whatwg.org/#dom-mutationobserver-takerecords)
- [IntersectionObserver#takeRecords](https://wicg.github.io/IntersectionObserver/#dom-intersectionobserver-takerecords)

PR #95, opened 2017-10-07
[WIP] Explain and improve PerformanceObserver ergonomics

igrigorik:
Per discussions [at the f2f](https://github.com/w3c/performance-timeline/issues/81#issuecomment-317089343) and WG calls, the high-level goals of this (WIP) update are:

- Improve developer ergonomics of PerformanceObserver
  - [x] define takeRecords interface to allow flushing of pending records - #74 
     - Addressed by https://github.com/w3c/performance-timeline/commit/c29d56980ba470dcb5f0a6d012c3e5ad18eefb8c. 
  - [ ] define static ~Performance#entryTypes to allow querying for supported types - #77.
  - [ ] define shared policy on buffering of PerformanceEntry objects before document onload event fires, to enable analytics consumers to retrieve emitted objects before they are able to register - #81.
- Define / refactor buffer management to avoid duplication and remove inconsistencies
  - [ ] align spec w/ existing implementations (separate buffers instead of global timeline buffer)
  - [ ] define ~registry mechanism that defines "supported entryTypes"
  - [ ] define reusable buffer management logic to reduce duplication / inconsistencies across specs

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/performance-timeline/perfobserver.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/performance-timeline/d4f4288...11c7e1c.html)


igrigorik:
PTAL at the "issue" in: https://s3.amazonaws.com/pr-preview/w3c/performance-timeline/perfobserver.html#entrytypes-attribute — before I get carried away here, would like to make sure we're still in agreement on the underlying plumbing and high-level approach.

/cc @toddreifsteck @JosephPecoraro @yoavweiss @spanicker

index.html
```diff
@@ -200,7 +200,18 @@
...
       </li>
+      <li>a list of <a>registered performance timeline emitter</a> objects. It is initially empty.</li>
     </ul>
+
+    <section>
+      <h2>Register <var>emitter</var> with <a>Performance Timeline</a></h2>
+      <p>An <var>emitter</var> represents a mechanism that can generate <a>PerformanceEntry</a> objects that are exposed to developers via the <a>Performance Timeline</a>. A <dfn>registered performance timeline emitter</dfn> consists of a:</p>
+      <ul>
+        <li><var>name</var> that MUST be unique.</li>
+        <li><var>performance entry buffer</var> that stores <a>PerformanceEntry</a> objects generated by the emitter, which is initially empty and has a <var>max-size</var> that, if unspecified, is set to 150.</li>
```
spanicker: At F2F we'd discussed buffering up to max buffer size OR load event? Not seeing any reference to load event here?

index.html
```diff
@@ -321,107 +334,138 @@
...
+
+          <p>We need to align our spec with implementation realities and also, ideally, define clear extension and reuse primitives to reduce duplication across specs. One way we could do this...</p>
+
+          <ul>
+            <li>Define a mechanism registry where upstream specs can register themselves with Peformance Timeline (this spec). The registration could consisnt of `name` and `size`, and Perf Timeline would then create the associated buffer and maintain it — run the add steps, enforce size constraints, etc.</li>
+            <li>Once we have the registry, we can also explain and define entryTypes — we have a list of registered names and that's what we'll return.</li>
```
spanicker: Isn't the set of supported types a static list?
Does the Custom Performance Entry proposal need registration requirement?
Could you clarify the need for dynamic registration?

index.html
spanicker: @tdresser as FYI - looks like Max's CL might be blocked on this being clarified in the spec.

igrigorik: @spanicker yep, good call, forgot to list this.. that's the behavior I was going to suggest.

- Each buffer has a max-size
- By default _all_ entries will be logged to their respective buffers up to onload event and size < max-size
- Some specs may also optionally choose to buffer after onload, at which point same size < max-size condition is enforced
  - if size =< max-size, new entries are dropped — i.e. we keep first N.

index.html
igrigorik: In theory, yes, it's a static list. In practice, however, different implementations enable support at different points in time, and hence we need to return a list of entry-type's _actually_ supported by the implementation to avoid [cases like this](https://github.com/w3c/performance-timeline/issues/87#issue-257140308).

Similarly, I don't want to freeze a list of strings in Perf Timeline: every time we add a new emitter, we'd have to update Perf Timeline spec, which would make things really weird and complicated for versioning. 

Last but not least, per our Blink conversation: imagine a future where certain emitters are only enabled if site is opted-in into some set of FP policies (e.g. no visited styling), that would require a further refinement in the algorithm that needs to account for what UA supports and the site policies.

tdresser: Regarding "Does the Custom Performance Entry proposal need registration requirement?" The [current proposal](https://docs.google.com/document/d/1_zm9JB-Ul_fOtAMnF6yBcmvNL3m4-k4ram4pdMIcIug/edit?pli=1#heading=h.ejti6qhmjv0b) suggests using an entryType of "custom" for all custom performance entry types, to avoid this problem.

igrigorik: @tdresser right, that works. All entries generated via `queueEntry()` would have the same entryType value, but we still need some mechanism to advertise that `custom` is observable — that's the registration I'm referring to here. Roughly, in Custom Entries spec this could look something like...

> Custom Performance Entries API participates in performance timeline and registers "custom" entry type and buffer size of 100 entries.

We could define "participates and registers" formally in Perf Timeline and then use that to plumb the rest.

tdresser: Right, that sounds good to me.

index.html
igrigorik: > Does the buffering logic impact when events are added to the performance timeline in any way? 

No, when an entry is queued is up to the upstream spec. 

> When we say buffer up to the onload event, what are we actually comparing?

I was thinking with respect to when the entry is queued, but entry.startTime could work too.

tdresser: I think when the entry is queued is correct, especially in the face of user timing entries with arbitrary start times.

SGTM.

index.html
```diff
@@ -321,107 +334,138 @@ <h2>The <dfn>PerformanceObserver</dfn> interface</h2>
...
+
+          <p>This spec defines the <a>performance entry buffer</a>, which we treat as a global buffer shared by all upstream emitters and which we query via `getEntries`, `getEntriesByName`, etc. <b>However</b>, in practice this is not how UA's implement this functionality. Instead of a global buffer, all implementations maintain separate buffers (each with own size limit), which we then query depending on name/type, etc. Further, some specs <b>strong</b> define own buffer management logic (e.g. ResourceTiming).</p>
+
+          <p>We need to align our spec with implementation realities and also, ideally, define clear extension and reuse primitives to reduce duplication across specs. One way we could do this...</p>
+
+          <ul>
+            <li>Define a mechanism registry where upstream specs can register themselves with Peformance Timeline (this spec). The registration could consisnt of `name` and `size`, and Perf Timeline would then create the associated buffer and maintain it — run the add steps, enforce size constraints, etc.</li>
```
tdresser: consist

igrigorik:
On closer review, I'll close this PR: we already merged most of the CLs here into main branch, and I want to separate entryTypes and buffer logic work. The respective bugs to track are:

- https://github.com/w3c/performance-timeline/issues/81 : buffering
- https://github.com/w3c/performance-timeline/issues/77 : entryTypes

Issue #94, opened 2017-10-04
Steps 3 and 4 of observe() aren't clear about what list is being considered


bzbarsky:
Step 3 of https://w3c.github.io/performance-timeline/#dom-performanceobserver-observe() says:

> If the list of registered performance observer objects associated with the ECMAScript global environment of the interface object's

I have no idea what "the interface object's" means here.  Please just explicitly refer to the current or relevant global as defined at <https://html.spec.whatwg.org/multipage/webappapis.html#realms-settings-objects-global-objects>, depending on which one you really want.  I believe Gecko uses the relevant global; can't speak for other implementations.

Note that I just checked, and there are no web platform tests that check which global is selected here, afaict.

There's a similar problem in step 4:

> append a new registered performance observer object to the list of registered performance observer objects associated with the ECMAScript global environment of the interface object

what "interface object"?  There isn't one around.

Issue #93, opened 2017-10-04
observe() step 3 is confused about what objects it's manipulating


bzbarsky:
https://w3c.github.io/performance-timeline/#dom-performanceobserver-observe() step 3 says:

> replace the context object's options with options. 

But the context object is a PerformanceObserver, and a PerformanceObserver doesn't have an `options` of any sort.

The thing that has an options is a "registered performance observer", which is presumably something like an infra tuple <https://infra.spec.whatwg.org/#tuple> or struct <https://infra.spec.whatwg.org/#struct>.  It would be better to explicitly spell out what sort of thing a "registered performance observer" is, and make sure that step 3 modifies the "registered performance observer", not the "context object".

Issue #91, opened 2017-10-04
Does PerformanceObserverInit's entryTypes contain types or names?


bzbarsky:
observe() algorithm steps talk about "names", but what it really contains is entry types, I assume.

Issue #90, opened 2017-10-04
Step 1 of observe() gets an entryTypes sequence from ... where?


bzbarsky:
Presumably this is meant to be the entryTypes member of the `options` dictionary, but the spec doesn't actually say that.  It would be better to be more explicit here.

PR #96, opened 2017-10-07
Improve observe() logic and references

igrigorik:
Integrate feedback from #90, #91, #93, #94.

@bzbarsky ptal.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/performance-timeline/observe-cleanup.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/performance-timeline/74a9181...6d7e857.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -549,7 +549,7 @@
       exception</dfn></li>
       <li><dfn data-cite="!WHATWG-HTML#task-queue">task queue</dfn></li>
       <li><dfn data-cite="!WHATWG-HTML#task-source">task source</dfn></li>
-      <li><dfn data-cite="!WHATWG-HTML#relevant-global-object">relevant global object</dfn></li>
+      <li><dfn data-cite="!WHATWG-HTML#concept-relevant-global">relevant global object</dfn></li>
       <li><dfn data-cite="!WHATWG-INFRA#struct">struct</dfn></li>
     </ul>
   </section>
--- a/index.html
+++ b/index.html
@@ -324,8 +324,8 @@
       <p>The <dfn>observe()</dfn> method instructs the user agent to
       <dfn>register the observer</dfn> and must run these steps:</p>
       <ol data-link-for="PerformanceObserverInit">
-        <li>Let <var>entry types</var> be <var>option</var>'s
-          <a for="PerformanceObserverInit">entryType</a> sequence.</li>
+        <li>Let <var>entry types</var> be <var>option</var>'s <a>entryTypes</a>
+          sequence.</li>
         <li>Remove all unsupported types from <var>entry types</var>. The user
           agent SHOULD notify developers if <var>entry types</var> is modified.
           For example, a console warning listing removed types might be appropriate.</li>
@@ -345,15 +345,15 @@
         interface object, with the <a>context object</a> as `observer` and
         <var>options</var> as the `options`.
         </li>
-        <li>If _options'_ <a for="PerformanceObserverInit">buffered</a> flag is
-          set, for each <dfn>entryType</dfn> of the `entryTypes` sequence:
+        <li>If _options'_ <a>buffered</a> flag is set, for each
+          <var>entry type</var> in <var>entry types</var> sequence:
           <ol>
-            <li>Let <dfn>entries</dfn> be the <a>PerformanceEntryList</a>
+            <li>Let <var>entries</var> be the <a>PerformanceEntryList</a>
             object returned by the
             <a href="#filter-buffer-by-name-and-type"></a> algorithm with
             <var>buffer</var> set to <a>performance entry buffer</a>,
             <var>name</var> set to `null` and <var>type</var> set to
-            <var>entryType</var>.
+            <var>entry type</var>.
             </li>
             <li>Append <var>entries</var> to the <a>context object</a>'s
             <a>observer buffer</a>.
@@ -379,7 +379,7 @@
         </pre>
         <dl>
           <dt><dfn>entryTypes</dfn></dt>
-          <dd>A list of entry names to be observed. The list MUST NOT be empty
+          <dd>A list of entry types to be observed. The list MUST NOT be empty
           and types not recognized by the user agent MUST be ignored.</dd>
         </dl>
         <dl>
--- a/index.html
+++ b/index.html
@@ -333,13 +333,13 @@
           steps are aborted to notify that registration has been aborted. For
           example, a console warning might be appropriate.</li>
         <li>If the <a>list of registered performance observer objects</a> of
-          <a>current global object</a> contains a <a>registered performance
+          <a>relevant global object</a> contains a <a>registered performance
           observer</a> that is the <a>context object</a>, replace its `options`,
           with <var>options</var>.
         </li>
        <li>Otherwise, append a new <a>registered performance observer</a>
         object to the <a>list of registered performance observer objects</a> of
-        <a>current global object</a>, with the <a>context object</a> as
+        <a>relevant global object</a>, with the <a>context object</a> as
         `observer` and <var>options</var> as the `options`.
         </li>
         <li>If <var>options</var>' <a>buffered</a> flag is set, for each
@@ -362,7 +362,7 @@
         <h2><dfn>disconnect()</dfn> method</h2>
         <p>The <a>disconnect()</a> method must remove the <a>context object</a>
         from the <a>list of registered performance observer objects</a> of
-        <a>current global object</a>, and also empty <a>context object</a>'s
+        <a>relevant global object</a>, and also empty <a>context object</a>'s
         <a>observer buffer</a>.</p>
       </section>
       <section data-dfn-for="PerformanceObserverInit" data-link-for=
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -324,33 +323,34 @@
...
+        <li>Remove all unsupported types from <var>entry types</var>. The user
+          agent SHOULD notify developers if <var>entry types</var> is modified.
+          For example, a console warning listing removed types might be appropriate.</li>
+        <li>If the resulting <var>entry types</var> sequence is an empty sequence,
+          abort these steps. The user agent SHOULD notify developers when the
+          steps are aborted to notify that registration has been aborted. For
+          example, a console warning might be appropriate.</li>
+        <li>If the <a>list of registered performance observer objects</a> of
+          <a>current global object</a> contains a <a>registered performance
+          observer</a> that is the <a>context object</a>, replace its `options`,
```
bzbarsky: Not possible.  The context object is a `PerformanceObserver`, but the list you're talking about contains "registered performance observer" instances, which is a different type.

This is really just https://github.com/w3c/performance-timeline/issues/93 remaining unfixed with the new text....

index.html
```diff
@@ -324,33 +323,34 @@
...
+          sequence.</li>
+        <li>Remove all unsupported types from <var>entry types</var>. The user
+          agent SHOULD notify developers if <var>entry types</var> is modified.
+          For example, a console warning listing removed types might be appropriate.</li>
+        <li>If the resulting <var>entry types</var> sequence is an empty sequence,
+          abort these steps. The user agent SHOULD notify developers when the
+          steps are aborted to notify that registration has been aborted. For
+          example, a console warning might be appropriate.</li>
+        <li>If the <a>list of registered performance observer objects</a> of
+          <a>current global object</a> contains a <a>registered performance
```
bzbarsky: Why current and not relevant?  What do implementations actually do here?

There's a pretty fundamental conceptual difference between using "current" and "relevant" here, unlike in some cases.  If "relevant" is used, then the observer can only observe things in one global (its "relevant global").  If "current" is used, then the same observer can observe things in multiple globals at once.  Which behavior are you aiming for?  If it's the "current global" behavior, it might be worth a note pointing out the multiple-global subtlety, because it would be easy to miss for both implementors and authors.

If you actually decide you want the "relevant global", I and probably @domenic would like to know why you ended up picking the "current global" here when that wasn't the thing you wanted.  Maybe we can make the HTML spec descriptions for these globals clearer somehow, if they're still confusing.

Either way, given the clear semantic difference above, this needs tests.

index.html
```diff
@@ -324,33 +323,34 @@
...
+          observer</a> that is the <a>context object</a>, replace its `options`,
+          with <var>options</var>.
         </li>
        <li>Otherwise, append a new <a>registered performance observer</a>
-        object to the list of <a>registered performance observer</a> objects
-        associated with the <a>ECMAScript global environment</a> of the
-        interface object, with the <a>context object</a> as `observer` and
-        <var>options</var> as the `options`.
+        object to the <a>list of registered performance observer objects</a> of
+        <a>current global object</a>, with the <a>context object</a> as
```
bzbarsky: Just make sure this matches the current vs relevant decision above.

index.html
```diff
@@ -324,33 +323,34 @@
...
-        <li>If _options'_ <a for="PerformanceObserverInit">buffered</a> flag is
-          set, for each <dfn>entryType</dfn> of the `entryTypes` sequence:
+        <li>If <var>options</var>' <a>buffered</a> flag is set, for each
+          <var>entry type</var> in <var>entry types</var> sequence:
           <ol>
-            <li>Let <dfn>entries</dfn> be the <a>PerformanceEntryList</a>
+            <li>Let <var>entries</var> be the <a>PerformanceEntryList</a>
             object returned by the
             <a href="#filter-buffer-by-name-and-type"></a> algorithm with
             <var>buffer</var> set to <a>performance entry buffer</a>,
```
bzbarsky: "performance entry buffer" for which global?  This needs to be specified.  And tested.

index.html
```diff
@@ -361,9 +361,9 @@
       <section>
         <h2><dfn>disconnect()</dfn> method</h2>
         <p>The <a>disconnect()</a> method must remove the <a>context object</a>
-        from the list of <a>PerformanceObserver</a> objects associated with the
-        <a>ECMAScript global environment</a> of the interface object, and also
-        empty <a>context object</a>'s <a>observer buffer</a>.</p>
+        from the <a>list of registered performance observer objects</a> of
+        <a>current global object</a>, and also empty <a>context object</a>'s
```
bzbarsky: Again, just make sure this matches the above current vs relevant decision.

index.html
```diff
@@ -465,8 +465,8 @@
           <li>Unset <a>performance observer task queued flag</a>.
```
bzbarsky: For which global?

index.html
```diff
@@ -465,8 +465,8 @@
...
           </li>
           <li>Let <i>notify list</i> be a copy of <a>ECMAScript global
```
bzbarsky: Which global?  It's not an input to this algorithm.  Needs to be specified.

index.html
igrigorik: Paging @spanicker @toddreifsteck @JosephPecoraro for implementation feedback.. :)

index.html
igrigorik: Thanks for flagging this, the mapping and differences were unclear in my head and staring at "8.1.3.5.3 Current" vs "8.1.3.5.4 Relevant" in HTML spec didn't help. 

Based on the differences you highlighted (which, btw, are a much approachable explainer that could probably make a useful note in the HTML spec, under either section), I believe I have this backwards: we want relevant. I'll update and we can confirm with other implementations.

index.html
igrigorik: Ack

igrigorik:
@bzbarsky thanks! Made a few updates, ptal.

index.html
```diff
@@ -193,14 +193,13 @@
...
-      empty.
-      </li>
+      <li>a <dfn>performance observer task queued flag</dfn></li>
+      <li>a <dfn>list of <a>registered performance observer</a> objects</dfn> that is initially empty</li>
       <li>a <dfn>performance entry buffer</dfn> to store
-      <a>PerformanceEntry</a> objects. It is initially empty.
+      <a>PerformanceEntry</a> objects that is initially empty
       </li>
     </ul>
+    <p>The <dfn>relevant performance entry buffer</dfn> is the <a>performance entry buffer</a> associated with the <a>relevant global object</a>.</p>
```
spanicker: I find this quite confusing.
There isn't a single buffer, but rather a buffer for each type of entry.
Should the spec reflect this?

index.html
```diff
@@ -221,22 +220,22 @@
         <h2><dfn>getEntries()</dfn> method</h2>
         <p>Returns a <a>PerformanceEntryList</a> object returned by
         <a href="#filter-buffer-by-name-and-type"></a> algorithm with
-        <var>buffer</var> set to <a>performance entry buffer</a>, and
+        <var>buffer</var> set to <a>relevant performance entry buffer</a>, and
```
spanicker: (same confusion with assuming single buffer) getEntries essentially looks at all the buffers and returns all of their entries

index.html
```diff
@@ -303,9 +302,9 @@
       <p>The `PerformanceObserver(callback)` constructor must create a new
       <a>PerformanceObserver</a> object with <a>PerformanceObserverCallback</a>
       set to <var>callback</var> and then return it.</p>
-      <p>A <dfn>registered performance observer</dfn> consists of an observer
-      (a <a>PerformanceObserver</a> object) and options (a
-      <a>PerformanceObserverInit</a> dictionary).</p>
+      <p>A <dfn>registered performance observer</dfn> is a <a>struct</a>
+      consisting of an <var>observer</var> (a <a>PerformanceObserver</a> object)
+      and <var>options</var> (a <a>PerformanceObserverInit</a> dictionary).</p>
```
spanicker: The options does not exit until observe is called, so this seems confusing,

index.html
igrigorik: Re, single buffer: yes, I think we should but that's out of scope for this PR. I flagged this very issue in https://github.com/w3c/performance-timeline/pull/95#issuecomment-336576176 — let's tackle that there. 

In the context of this PR, let's make sure we nail down the correct definition for globals. 

Does `relevant global object` look correct to you?

index.html
igrigorik: Yep, and we have the plumbing in place in observe. Here we're simply defining what a "registered performance observer" is.

index.html
bzbarsky: So this has nothing to do with the "each ECMAScript global environment" bit right before it, right?

"relevant global object" is a broken link in the rendered version of the spec...  I _assume_ you meant https://html.spec.whatwg.org/multipage/#concept-relevant-global but note that this needs as input the specific platform object you're talking about.  This comment about "relevant global object" applies to other places in this PR too..

index.html
bzbarsky: So here, you'd probably want "relevant performance entry buffer for the context object" or some such.  Or you could define "relevant performance entry buffer", when used in an algorithm, to explicitly use the relevant global of https://dom.spec.whatwg.org/#context-object for that algorithm.

igrigorik: @bzbarsky yep, that's what we have on line 202.

index.html
igrigorik: > So this has nothing to do with the "each ECMAScript global environment" bit right before it, right?

Correct. This definition is separate from the list defined above.

> I assume you meant https://html.spec.whatwg.org/multipage/#concept-relevant-global but note that this needs as input the specific platform object you're talking about. 

Great catch, thank you. Fixed.

index.html
igrigorik: Yep, that's exactly what we have in 202 and reference here.

PR #98, opened 2017-12-08
define PerformanceObserver#takeRecords()

### LINKED ISSUE #74: Add support for flushing observer via `takeRecords()`

igrigorik:
Closes #74.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/pull/98.html" title="Last updated on Dec 8, 2017, 10:37 PM GMT (9154b82)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/98/43cf7a3...9154b82.html" title="Last updated on Dec 8, 2017, 10:37 PM GMT (9154b82)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -312,6 +312,7 @@
       interface PerformanceObserver {
         void observe (PerformanceObserverInit options);
         void disconnect ();
+        PerformanceEntryList takeRecords();
       };
       </pre>
       <p class="note">To keep the performance overhead to minimum the
@@ -415,6 +416,15 @@
             to `type` otherwise.</p>
           </section>
         </section>
+      </section>
+      <section>
+        <h2><dfn>takeRecords()</dfn> method</h2>
+        <p>The <a>takeRecords()</a> method instructs the user agent to flush  entries in <a>context object</a>'s <a>observer buffer</a> and must run these steps:</p>
+        <ol data-link-for="PerformanceObserver">
+          <li>Let <var>entries</var> be a copy of <a>context object</a>'s <a>observer buffer</a>.</li>
+          <li>Empty <a>context object</a>'s <a>observer buffer</a>.</li>
+          <li>Return <var>entries</var>.</li>
+        </ol>
       </section>
       <section>
         <h2><dfn>disconnect()</dfn> method</h2>
--- a/index.html
+++ b/index.html
@@ -295,7 +295,7 @@
       <p>Each <a>PerformanceObserver</a> has these associated concepts:</p>
       <ul>
         <li>A <dfn>PerformanceObserverCallback</dfn> set on creation.</li>
-        <li>A list of <a>PerformanceEntry</a> objects called the <dfn>observer
+        <li> A <a>PerformanceEntryList</a> object called the <dfn>observer
         buffer</dfn> that is initially empty.
         </li>
       </ul>
@@ -419,12 +419,9 @@
       </section>
       <section>
         <h2><dfn>takeRecords()</dfn> method</h2>
-        <p>The <a>takeRecords()</a> method instructs the user agent to flush  entries in <a>context object</a>'s <a>observer buffer</a> and must run these steps:</p>
-        <ol data-link-for="PerformanceObserver">
-          <li>Let <var>entries</var> be a copy of <a>context object</a>'s <a>observer buffer</a>.</li>
-          <li>Empty <a>context object</a>'s <a>observer buffer</a>.</li>
-          <li>Return <var>entries</var>.</li>
-        </ol>
+        <p>The <a>takeRecords()</a> method must return a copy of the
+          <a>context object</a>'s <a>observer buffer</a>, and also empty
+          <a>context object</a>'s <a>observer buffer</a>.</p>
       </section>
       <section>
         <h2><dfn>disconnect()</dfn> method</h2>


igrigorik:
@npm1 we also need to add WPT tests for this update, do you have anything inflight?

index.html
```diff
@@ -416,6 +417,15 @@ <h2><dfn>getEntriesByName()</dfn> method</h2>
           </section>
         </section>
       </section>
+      <section>
+        <h2><dfn>takeRecords()</dfn> method</h2>
+        <p>The <a>takeRecords()</a> method instructs the user agent to flush  entries in <a>context object</a>'s <a>observer buffer</a> and must run these steps:</p>
+        <ol data-link-for="PerformanceObserver">
+          <li>Let <var>entries</var> be a copy of <a>context object</a>'s <a>observer buffer</a>.</li>
+          <li>Empty <a>context object</a>'s <a>observer buffer</a>.</li>
+          <li>Return <var>entries</var>.</li>
```
tdresser: Don't we need to convert the list of PerformanceEntry objects to a PerformanceEntryList?

index.html
```diff
@@ -416,6 +417,15 @@ <h2><dfn>getEntriesByName()</dfn> method</h2>
           </section>
         </section>
       </section>
+      <section>
+        <h2><dfn>takeRecords()</dfn> method</h2>
+        <p>The <a>takeRecords()</a> method instructs the user agent to flush  entries in <a>context object</a>'s <a>observer buffer</a> and must run these steps:</p>
```
tdresser: flush  entries -> flush entries

in <a>context object -> in the <a>context object

I'm a bit confused by this.
Is this equivalent to saying 
"and must run these steps" -> "by running these steps"?

igrigorik: Ack. 

> "and must run these steps" -> "by running these steps"?

The preamble is short explainer for the method, and the algorithm are the actual instructions that the UA must execute; we use this pattern in other other places in the spec.

npm1: I agree that "by running these steps" is clearer. We could modify those other places in a followup.

npm1:
We do not currently have takeRecords tests, so I'll need to add some. Shouldn't be hard to do.

index.html
igrigorik: Those are equivalent "typedef sequence<PerformanceEntry> PerformanceEntryList;"

That said, agree that this should be clearer, will update the definition.

igrigorik:
@tdresser @npm1 updated and simplified to align with other definitions.

index.html
tdresser: Oops, that makes sense, thanks!

index.html
```diff
@@ -416,6 +417,12 @@ <h2><dfn>getEntriesByName()</dfn> method</h2>
           </section>
         </section>
       </section>
+      <section>
+        <h2><dfn>takeRecords()</dfn> method</h2>
+        <p>The <a>takeRecords()</a> method must return a copy of the
+          <a>context object</a>'s <a>observer buffer</a>, and also empty
```
tdresser: Let me know if this is being overly picky.

I'd prefer "and then empty" vs "and also empty", as "also" doesn't as strongly imply an order to these operations.

index.html
```diff
@@ -416,6 +417,12 @@ <h2><dfn>getEntriesByName()</dfn> method</h2>
...
+          <a>context object</a>'s <a>observer buffer</a>.</p>
```
tdresser: the \<a>context ...

npm1:
lgtm

npm1:
I have a question that might be worth clarifying in the spec: what happens when takeRecords() is called during a callback of the observer?

npm1:
Ping, is this ready to go?

toddreifsteck:
@rniwa @wanderview @JosephPecoraro  are any of you able to take a look and ensure there is nothing you are fundamentally against here?

I have no pushback on the concept.

rniwa:
Seems fine assuming this works like `MutationObserver`'s `takeRecords`.

igrigorik:
> I have a question that might be worth clarifying in the spec: what happens when takeRecords() is called during a callback of the observer?

Practically speaking, I think this will be a no-op in most cases (as in, it'll return an empty entrilyst) as long as the callback is not taking its sweet time to complete: the entrylist passed to the callback should already have the latest copy of everything in the observer's buffer. That said, I guess it is possible that in the time delta between when observer's callback receives its list and when takeRecords is called, some new entry is queued.. in that case, the observer's buffer is flushed synchronously and will return what's in the buffer. 

As far as processing logic, I don't think there is anything special here? Let me know if otherwise.

> @rniwa: Seems fine assuming this works like MutationObserver's takeRecords.

Yep, that's the intent.

---

@toddreifsteck @rniwa @tdresser thanks for the reviews. I think this is good to merge as soon as we address the last question by @npm1.

npm1:
I don't recall if I had any special case in mind. Looking at the descriptions, it's clear that takeRecords will return nil (or any entries observed after starting the callback). So still LGTM.

igrigorik:
Great, then I think we're good to merge. Thanks everyone!

Issue #103, opened 2018-07-23
PerformanceObserver's observe() should support entryType-specific parameters


npm1:
Currently, observe() gets a sequence of entryTypes and optionally a boolean buffered flag. I'd like to augment this as follows:

* Keep the entryTypes sequence.
* Allow an optional dictionary for each entryType that can contain entryType-specific parameters.
* Move the buffered flag to become part of the entryType-specific dictionaries. I believe no browser vendor has implemented this flag yet.

The entryTypes could be used alongside entryType-specific parameters. For example, a user could say:
`observer.observe({entryTypes: ["resource"],
longtask: {buffered: true, threshold: 100}
});`

rniwa:
So the whole *Observer pattern is that you're supposed to call `observe` on each thing you observe. So it's not great that we're making observe function take multiple things to observe just in `PerformanceObserver`.

I think a better interface would be to make observe(~) take "type" and other options directly as in:
```js
observe.observe({entryTypes: ["resource"]});
observe.observe({type: 'longtask', buffered: true, threshold: 100});
```

npm1:
We already allow observe to take multiple things to observe. And if there are multiple calls to observe() I think that currently we override what we observe, not append to it.

igrigorik:
Hmm.. Our earlier proposal was predicated on assumption that folks want to use a single observer to monitor for multiple types, each with potentially different settings. On the WG call, I think we identified that this may not be the best pattern, or pattern we want to encourage. Instead of observing many different types and building a switch statement, create multiple distinct observers. 

Given that, WDYT of...

```javascript
observe.observe({entryTypes: ["resource", "navigation"], buffered: true, threshold: 100});
```

Observe `resource` and `navigation` with same `buffered` and `threshold` settings applying to both. 

```javascript
observe.observe({entryTypes: ["mark"], buffered: true});
observe.observe({entryTypes: ["longtask"], buffered: true, threshold: 50});
```
If you want to observe different types with different settings, create multiple observers? Perhaps we can just stick with `entryTypes`? Conceptually it's pretty simple to grok and explain, and if we stick with same attributes, we don't need to create any special precedence rules for processing.

npm1:
It seems to me that allowing a single observe() call to specify multiple entryTypes and also various settings is prone to errors if the names of the settings are shared but they mean different things for different entryTypes. Thus, a developer could do:
`observer.observe({entryTypes: ["resource", "longtask"], buffered: true, threshold: 100});`
[...]

PR #104, opened 2018-07-24
Add entryType-specific parameters for observe()

npm1:
This PR adds a PerformanceEntryObserverOptions dictionary where entryType-specific parameters are used. The buffered flag is moved to this dictionary. Some \<i\> tags are changed to \<var\>. Observe() is updated to consider dictionaries. Queueing an entry is modified to allow passing in a closure. This is in preparation for adding thresholds to certain entry types. Solves https://github.com/w3c/performance-timeline/issues/103
@igrigorik @tdresser @toddreifsteck PTAL

***
<a href="https://pr-preview.s3.amazonaws.com/npm1/performance-timeline/pull/104.html" title="Last updated on Sep 7, 2018, 9:48 PM GMT (c4f8432)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/104/81a6242...npm1:c4f8432.html" title="Last updated on Sep 7, 2018, 9:48 PM GMT (c4f8432)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -383,12 +383,19 @@
           <pre class="idl">
           dictionary PerformanceObserverInit {
             required sequence&lt;DOMString&gt; entryTypes;
+            PerformanceEntryObserverOptions fallback;
           };
           </pre>
           <dl>
             <dt><dfn>entryTypes</dfn></dt>
             <dd>A list of entry types to be observed. The list MUST NOT be empty
             and types not recognized by the user agent MUST be ignored.</dd>
+            <dt><dfn>fallback</dfn></dt>
+            <dd>A dictionary containing the parameters to be applied to all entries
+            observed. These parameters are overriden by entryType-specific dictionaries.
+            That is, if a parameter is present in the `fallback` dictionary but also in
+            the `mark` dictionary, then the value of the `mark` dictionary will be used
+            for entries whose <a data-lt="PerformanceEntry.entryType">entryType</a> is `mark`.</dd>
           </dl>
           <p class="note">Whenever a new kind of <a>PerformanceEntry</a> is defined,
             the <a>PerformanceObserverInit</a> dictionary should be augmented to
--- a/index.html
+++ b/index.html
@@ -300,6 +300,9 @@
         <li> A <a>PerformanceEntryList</a> object called the <dfn>observer
         buffer</dfn> that is initially empty.
         </li>
+        <li>A sequence of strings called <dfn>supported entry types</dfn> that is
+          initially empty.
+        </li>
       </ul>
       <p>The `PerformanceObserver(callback)` constructor must create a new
       <a>PerformanceObserver</a> object with <a>PerformanceObserverCallback</a>
@@ -328,27 +331,33 @@
         <p>The <a>observe()</a> method instructs the user agent to
         <dfn>register the observer</dfn> and must run these steps:</p>
         <ol data-link-for="PerformanceObserverInit">
-          <li>Let <var>entry types</var> be <var>options</var>' <a>entryTypes</a>
-            sequence.</li>
-          <li>For each dictionary of name <var>entryType</var> in <var>options</var>,
-            add the string `<var>entryType</var>` to <var>entry types</var>.
-          <li>Remove all unsupported types from <var>entry types</var>. The user
-            agent SHOULD notify developers if <var>entry types</var> is modified.
-            For example, a console warning listing removed types might be appropriate.</li>
-          <li>If the resulting <var>entry types</var> sequence is an empty sequence,
+          <li>Set <a>supported entry types</a> to a copy of <var>options</var>'
+            <a>entryTypes</a> sequence.
+          </li>
+          <li>For each key <var>entryType</var> in the <var>options</var> dictonary that
+            is not <a>entryTypes</a>, add the string <var>entryType</var> to
+            <a>supported entry types</a>.
+          </li>
+          <li>Remove all unsupported types from <a>supported entry types</a>. The user
+            agent SHOULD notify developers if <a>supported entry types</a> is modified.
+            For example, a console warning listing removed types might be appropriate.
+          </li>
+          <li>If <a>supported entry types</a> is an empty sequence,
             abort these steps. The user agent SHOULD notify developers when the
             steps are aborted to notify that registration has been aborted. For
-            example, a console warning might be appropriate.</li>
+            example, a console warning might be appropriate.
+          </li>
           <li>If the <a>list of registered performance observer objects</a> of
             <a>relevant global object</a> contains a <a>registered performance
-            observer</a> whose `observer` is the <a>context object</a>, replace its `options`, with <var>options</var>.
+            observer</a> whose `observer` is the <a>context object</a>, replace its
+            `options`, with <var>options</var>.
           </li>
          <li>Otherwise, append a new <a>registered performance observer</a>
           object to the <a>list of registered performance observer objects</a> of
           <a>relevant global object</a>, with the <a>context object</a> as
           `observer` and <var>options</var> as the `options`.
           </li>
-          <li>For each <var>entry type</var> in <var>entry types</var>, if the
+          <li>For each <var>entry type</var> in <a>supported entry types</a>, if the
             <a data-link-for="PerformanceEntryObserverOptions">buffered</a> flag is set
             in <var>options</var>' dictionary named <var>entry type</var>:
             <ol>
@@ -470,12 +479,11 @@
         conditions are met:
         <ol>
           <li><var>new entry</var>’s <a data-lt="PerformanceEntry.entryType">entryType</a>
-          value is in <var>observer</var>'s <a>PerformanceObserverInit</a>: either its
-          <a data-lt="PerformanceObserverInit.entryTypes">entryTypes</a> includes it or it
-          has a dictionary with that name.</li>
+            value is in <var>observer</var>'s <a>supported entry types</a>.
+          </li>
           <li><var>should append to observer</var>(<var>new entry</var>,
             <var>observer</var>) returns true, if defined by the <i>specification</i> of
-            the corresponding <a data-lt="PerformanceEntry.entryType">entryType</a>.</li>
+            the corresponding <a data-lt="PerformanceEntry.entryType">entryType</a>.
           </li>
         </ol>
       </li>
--- a/index.html
+++ b/index.html
@@ -451,14 +451,13 @@
     <h2>Processing</h2>
     <section data-link-for="PerformanceObserver">
     <h2>Queue a <code>PerformanceEntry</code></h2>
-      <p>The <dfn>queue a PerformanceEntry</dfn> algorithm receives these parameters:</p>
-      <li>A PerformanceEntry <var>new entry</var>.</li>
-      <li>An optional <var>add to performance entry buffer</var> flag, which is unset
-      by default.</li>
-      <li>An optional <var>should append to observer</var> closure whose input is a
-        <a>PerformanceObserver</a> and whose output is a boolean. By default this is the
-        closure which always returns true.</li>
-      <p>The algorithm runs these steps:</p>
+      <p><i>Specifications</i> of performance entries may define an algorithm
+        <var>should append to observer</var> which receives a <a>PerformanceEntry</a> and
+        a <a>PerformanceObserver</a> as input and returns a boolean specifying whether the
+        observer should be notified of the entry.
+      <p>The <dfn>queue a PerformanceEntry</dfn> <var>new entry</var> with an optional
+        <var>add to performance entry buffer</var> flag, which is unset by default, run
+        these steps:</p>
     <ol>
       <li>Let <var>interested observers</var> be an initially empty set of
       <a>PerformanceObserver</a> objects.
@@ -471,7 +470,9 @@
           value is in <var>observer</var>'s <a>PerformanceObserverInit</a>: either its
           <a data-lt="PerformanceObserverInit.entryTypes">entryTypes</a> includes it or it
           has a dictionary with that name.</li>
-          <li><var>should append to observer</var>(<var>observer</var>) returns true.</li>
+          <li><var>should append to observer</var>(<var>new entry</var>,
+            <var>observer</var>) returns true, if defined by the <i>specification</i> of
+            the corresponding <a data-lt="PerformanceEntry.entryType">entryType</a>.</li>
           </li>
         </ol>
       </li>
--- a/index.html
+++ b/index.html
@@ -376,10 +376,11 @@
             <dd>A list of entry types to be observed. The list MUST NOT be empty
             and types not recognized by the user agent MUST be ignored.</dd>
           </dl>
-          <p class="note">Whenever a new kind of <a>PerformanceEntry</a> is defined, this
-          dictionary should be augmented to include a new optional
-          <a>PerformanceEntryObserverOptions</a> dictionary with the name of the new
-          <a data-lt="PerformanceEntry.entryType">entryType</a>.</p>
+          <p class="note">Whenever a new kind of <a>PerformanceEntry</a> is defined,
+            the <a>PerformanceObserverInit</a> dictionary should be augmented to
+            include a new optional <a>PerformanceEntryObserverOptions</a> dictionary
+            with the name of the new <a data-lt="PerformanceEntry.entryType">
+            entryType</a>.</p>
         </section>
         <section data-dfn-for="PerformanceEntryObserverOptions" data-link-for=
         "PerformanceEntryObserverOptions">
@@ -455,9 +456,9 @@
         <var>should append to observer</var> which receives a <a>PerformanceEntry</a> and
         a <a>PerformanceObserver</a> as input and returns a boolean specifying whether the
         observer should be notified of the entry.
-      <p>The <dfn>queue a PerformanceEntry</dfn> <var>new entry</var> with an optional
-        <var>add to performance entry buffer</var> flag, which is unset by default, run
-        these steps:</p>
+      <p>To <dfn>queue a PerformanceEntry</dfn> (<var>new entry</var>) with an
+        optional <var>add to performance entry buffer</var> flag, which is unset by
+        default, run these steps:</p>
     <ol>
       <li>Let <var>interested observers</var> be an initially empty set of
       <a>PerformanceObserver</a> objects.
--- a/index.html
+++ b/index.html
@@ -151,13 +151,15 @@
           };
           return JSON.stringify(obj, null, 2);
         })
-        // Display them to the console
+        // Display them to the console.
         .forEach(console.log);
-      // maybe disconnect after processing the events.
+      // Maybe disconnect after processing the events.
       observer.disconnect();
     });
-    // subscribe to new events for Resource-Timing and User-Timing
-    // and retrieve buffered events from Resource-Timing 
+    // Subscribe to new events for User-Timing (marks and measures) via the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -366,14 +369,26 @@ <h2><dfn>PerformanceObserverInit</dfn> dictionary</h2>
...
-            boolean buffered = false;
           };
           </pre>
           <dl>
             <dt><dfn>entryTypes</dfn></dt>
             <dd>A list of entry types to be observed. The list MUST NOT be empty
             and types not recognized by the user agent MUST be ignored.</dd>
           </dl>
+          <p class="note">Whenever a new kind of <a>PerformanceEntry</a> is defined, this
+          dictionary should be augmented to include a new optional
```
tdresser: What is "this dictionary"?

index.html
```diff
@@ -436,30 +451,39 @@ <h2><dfn>disconnect()</dfn> method</h2>
...
     <section data-link-for="PerformanceObserver">
     <h2>Queue a <code>PerformanceEntry</code></h2>
-      <p>To <dfn>queue a PerformanceEntry</dfn> (<var>new entry</var>) with an
-      optional <var>add to performance entry buffer</var> flag, which is unset
-      by default, run these steps:</p>
+      <p><i>Specifications</i> of performance entries may define an algorithm
+        <var>should append to observer</var> which receives a <a>PerformanceEntry</a> and
+        a <a>PerformanceObserver</a> as input and returns a boolean specifying whether the
+        observer should be notified of the entry.
+      <p>The <dfn>queue a PerformanceEntry</dfn> <var>new entry</var> with an optional
```
tdresser: I'm not following the grammar of this sentence.

index.html
```diff
@@ -436,30 +451,39 @@ <h2><dfn>disconnect()</dfn> method</h2>
...
-          <li>If <i>observer</i>'s <a>PerformanceObserverInit</a> <a data-lt=
-          "PerformanceObserverInit.entryTypes">entryTypes</a> includes <i>new
-          entry</i>’s <a data-lt="PerformanceEntry.entryType">entryType</a>
-          value, append <i>observer</i> to <i>interested observers</i>.
+          <li><var>new entry</var>’s <a data-lt="PerformanceEntry.entryType">entryType</a>
+          value is in <var>observer</var>'s <a>PerformanceObserverInit</a>: either its
+          <a data-lt="PerformanceObserverInit.entryTypes">entryTypes</a> includes it or it
+          has a dictionary with that name.</li>
+          <li><var>should append to observer</var>(<var>new entry</var>,
+            <var>observer</var>) returns true, if defined by the <i>specification</i> of
```
tdresser: For cloning steps, I thought there was an inheritance like approach taken, where specs could override the default. That feels a bit slicker than what we've got here.

index.html
npm1: This is in the section titled 'PerformanceObserverInit dictionary' so I think it's clear enough? To reduce ambiguity I could replace that with 'the PerformanceObserverInit dictionary'.

index.html
npm1: Sorry, which? The one starting with Specifications, or with 'The queue'

index.html
npm1: Hm no, I don't think there is a default. The closure approach has a default, the cloning steps approach says that a specification may specify additional steps that should be run at a certain point in time.

index.html
npm1: Definitely The queue sentence is incorrect. The first word should be 'To'.

index.html
tdresser: You're right!

npm1:
Ping Ilya and Todd (or who else could be a good reviewer for this?)

index.html
```diff
@@ -156,11 +156,11 @@ <h2>Introduction</h2>
       // maybe disconnect after processing the events.
       observer.disconnect();
     });
-    // retrieve buffered events and subscribe to new events
-    // for Resource-Timing and User-Timing
+    // subscribe to new events for Resource-Timing and User-Timing
+    // and retrieve buffered events from Resource-Timing
```
igrigorik: It's probably worth spelling out explicitly what's going on here. Specially, that providing names to `entryTypes` enables monitoring for said type, initialized with some default settings (e.g. buffered = false). Alternatively, you can specify the entryType name as a key and provide a dictionary with custom settings — e.g. `resource : {buffered: true}`. 

On that note, @toddreifsteck @yoavweiss would love to get your sanity check and review on this pattern.

index.html
```diff
@@ -326,8 +326,10 @@ <h2><dfn>observe()</dfn> method</h2>
         <p>The <a>observe()</a> method instructs the user agent to
         <dfn>register the observer</dfn> and must run these steps:</p>
         <ol data-link-for="PerformanceObserverInit">
-          <li>Let <var>entry types</var> be <var>options</var> <a>entryTypes</a>
+          <li>Let <var>entry types</var> be <var>options</var>' <a>entryTypes</a>
             sequence.</li>
+          <li>For each dictionary of name <var>entryType</var> in <var>options</var>,
+            add the string `<var>entryType</var>` to <var>entry types</var>.
```
igrigorik: Not sure I grok what this is doing. What is an example of "each dictionary of name `entryType`"? :-)

I _think_ you're enumerating the keys in options and adding them to entryTypes? As an aside, can we actually do this without first cloning the options? As in, I don't think we should be modifying the passed in object.

index.html
```diff
@@ -366,14 +369,27 @@ <h2><dfn>PerformanceObserverInit</dfn> dictionary</h2>
...
           <dl>
             <dt><dfn>entryTypes</dfn></dt>
             <dd>A list of entry types to be observed. The list MUST NOT be empty
             and types not recognized by the user agent MUST be ignored.</dd>
           </dl>
+          <p class="note">Whenever a new kind of <a>PerformanceEntry</a> is defined,
+            the <a>PerformanceObserverInit</a> dictionary should be augmented to
+            include a new optional <a>PerformanceEntryObserverOptions</a> dictionary
+            with the name of the new <a data-lt="PerformanceEntry.entryType">
+            entryType</a>.</p>
```
igrigorik: We don't do this today for existing types. If we follow this recommendation, than we should do so.

That said, we've also tried to explicitly avoid enumerating "valid types" in Performance Timeline, as that makes a hard dependency for the perf timeline to be revved whenever we add a new type. Per discussion in https://github.com/w3c/performance-timeline/issues/77, the idea is to have other specs call into Perf Timeline to register available types. The logic here should then enumerate the dictionary members, compare against registered + known types and proceed accordingly. This also means that we ignore any unknown types in this algorithm.

WDYT, does that make sense?

index.html
```diff
@@ -436,30 +452,39 @@ <h2><dfn>disconnect()</dfn> method</h2>
     <h2>Processing</h2>
     <section data-link-for="PerformanceObserver">
     <h2>Queue a <code>PerformanceEntry</code></h2>
+      <p><i>Specifications</i> of performance entries may define an algorithm
```
igrigorik: I believe our plan was to have this on by default for all types prior to onload, and no buffering afterwards. If that's the case, I believe we wouldn't need this — correct?

index.html
```diff
@@ -436,30 +452,39 @@ <h2><dfn>disconnect()</dfn> method</h2>
...
+        <var>should append to observer</var> which receives a <a>PerformanceEntry</a> and
+        a <a>PerformanceObserver</a> as input and returns a boolean specifying whether the
+        observer should be notified of the entry.
       <p>To <dfn>queue a PerformanceEntry</dfn> (<var>new entry</var>) with an
-      optional <var>add to performance entry buffer</var> flag, which is unset
-      by default, run these steps:</p>
+        optional <var>add to performance entry buffer</var> flag, which is unset by
+        default, run these steps:</p>
```
igrigorik: This step should be with respect to before or after onload, not the passed in flag. 

For context, we ruled out the flag approach because it would, once again, require updating every upstream spec to have this on by default.. and I can't think of cases where we'd want this to be off by default for before-onload.

index.html
```diff
@@ -436,30 +452,39 @@ <h2><dfn>disconnect()</dfn> method</h2>
...
+        conditions are met:
         <ol>
-          <li>If <i>observer</i>'s <a>PerformanceObserverInit</a> <a data-lt=
-          "PerformanceObserverInit.entryTypes">entryTypes</a> includes <i>new
-          entry</i>’s <a data-lt="PerformanceEntry.entryType">entryType</a>
-          value, append <i>observer</i> to <i>interested observers</i>.
+          <li><var>new entry</var>’s <a data-lt="PerformanceEntry.entryType">entryType</a>
+          value is in <var>observer</var>'s <a>PerformanceObserverInit</a>: either its
+          <a data-lt="PerformanceObserverInit.entryTypes">entryTypes</a> includes it or it
+          has a dictionary with that name.</li>
```
igrigorik: Why are we doing this here again, if we're modifying entryTypes and adding names to it earlier? Could we preprocess this earlier (either unroll entryTypes into the array with default dicts, or the reverse) and reuse that here?

index.html
npm1: Yes, maybe enumarating keys in options is clearer? Or enumerating keys whose values are dictionaries (to exclude entryTypes itself for example)?

index.html
npm1: I'm not forcing PerformanceTimeline to be modified whenever a new entry type is added. I was thinking the new spec could do something along the lines of 'partial interface' to add the new dictionary corresponding to its entry type. We already "remove all unsupported types", which is a bit hand-wavy but I think it's clear what it means. If we want to make it a bit more formal, we can go with your registration approach, I just don't know if it adds much clarity or if it makes speccing new entry types easier.

index.html
npm1: This is about the queue algorithm, not about buffering. An example of where we need this is LongTasks thresholds: we need the spec to provide an algorithm to determine if an entry should be added to an observer (look at the entry's duration, compare with the observer's threshold).

index.html
npm1: Sure, but I haven't added this flag, it's currently in the spec. How about we do a separate PR for this independent change?

index.html
npm1: We did this in observe() to the |options| parameter, and now that is gone. I would propose having an internal variable within PerformanceObserver that consists of a list of supported entry types. The list could be populated during observe() and then it could be used here, in the queue an entry algorithm. This is probably how implementations do it anyways. WDYT?

index.html
igrigorik: Per later comment, I don't think we should be modifying this object in place.. But if we're cloning it first, we could: pluck entryTypes out first and `<do-stuff>`, then enumerate the remaining keys and `<do-more-stuff>`, arriving at a consistent internal object that we can use later in the algorithm?

npm1: Yes, so I propose adding something like "Each PerformanceObserver gets an associated |entry types| object". Then here we would modify observer's |entry types| instead of some variable of option's entryTypes array. Does that sound reasonable?

index.html
```diff
@@ -366,14 +380,27 @@ <h2><dfn>PerformanceObserverInit</dfn> dictionary</h2>
           <pre class="idl">
           dictionary PerformanceObserverInit {
             required sequence&lt;DOMString&gt; entryTypes;
```
npm1: I'm thinking we may want to remove the 'required' from |entryTypes| as well as you could specify what you want to observe using the other dictionaries.

npm1:
By the way I've added the 'supported entry types' and addressed all comments.
@yoavweiss @toddreifsteck any thoughts on the pattern used on observe() here?

index.html
igrigorik: Ah, I'm with you now. I like that, that's much cleaner than I what I had in mind before. The one bit I'd change in this proposal is, instead of extending the `PerformanceObserverInit` we wanted + need to introduce `performance.entryTypes` and that's the interface the various specs should be extending. In turn, we can then filter passed options to `PerformanceObserverInit` against said interface.

WDYT, does that make sense? If you're on board with that.. the changes are coupled, and we would need to land the `performance.entryTypes` proposal first, and then rebase this against that.

index.html
```diff
@@ -298,6 +300,9 @@ <h2>The <dfn>PerformanceObserver</dfn> interface</h2>
         <li> A <a>PerformanceEntryList</a> object called the <dfn>observer
         buffer</dfn> that is initially empty.
         </li>
+        <li>A sequence of strings called <dfn>supported entry types</dfn> that is
```
igrigorik: s/supported entry types/observed entry types/? 

I think we should reserve `supported entry types` for the global "which entry types does performance timeline support"?

index.html
npm1: Sorry, I don't follow what you're suggesting here... `performance.entryTypes` will only help in knowing which are the entry types that are supported globally (implemented by the browser), right? So we still need a way to specify what filters can be used for each entry type, unless we want to support the same set of filters for all entry types? Even if we have the same set of filters on all entry types, we still need to augment `PerformanceObserverInit` with the dictionaries per entry type (I don't think we want to allow it to contain dictionaries like 'longtask' and 'mark' without specifying them anywhere explicitly). The only alternative I can think of that avoids this would be to have something like this:
```
  required sequence<DOMString> entryTypes;
  PerformanceEntryObserverOptions filters;
}
```
But this would force an observer to apply the same set of filters to all of the observed entry types that it is observing. This is kind of what we're trying to avoid here, since ideally we'd like entryType-specific parameters. If we don't have that, we might as well just keep adding the filters to `PerformanceObserverInit` directly. WDYT?

index.html

index.html
```diff
@@ -326,26 +331,35 @@ <h2><dfn>observe()</dfn> method</h2>
         <p>The <a>observe()</a> method instructs the user agent to
         <dfn>register the observer</dfn> and must run these steps:</p>
         <ol data-link-for="PerformanceObserverInit">
-          <li>Let <var>entry types</var> be <var>options</var> <a>entryTypes</a>
-            sequence.</li>
-          <li>Remove all unsupported types from <var>entry types</var>. The user
-            agent SHOULD notify developers if <var>entry types</var> is modified.
-            For example, a console warning listing removed types might be appropriate.</li>
-          <li>If the resulting <var>entry types</var> sequence is an empty sequence,
+          <li>Set <a>observed entry types</a> to a copy of <var>options</var>'
```
yoavweiss: I think it'd be better to explicitly state what the options variable is a line above that.
e.g. "Let <var>options</var> be the input <a>PerformanceObserverInit</a> object" or something along those lines

yoavweiss: Also, I think you meant "observed entry types" to be a var, not a link

index.html
```diff
@@ -326,26 +331,35 @@ <h2><dfn>observe()</dfn> method</h2>
...
+            <a>entryTypes</a> sequence.
+          </li>
+          <li>For each key <var>entryType</var> in the <var>options</var> dictonary that
```
yoavweiss: s/dictonary/dictionary/

index.html
```diff
@@ -326,26 +331,35 @@ <h2><dfn>observe()</dfn> method</h2>
...
+            is not <a>entryTypes</a>, add the string <var>entryType</var> to
+            <a>observed entry types</a>.
```
yoavweiss: It's not clear to me what this does and why. Can you explain?

index.html
```diff
@@ -326,26 +331,35 @@ <h2><dfn>observe()</dfn> method</h2>
...
+          </li>
+          <li>Remove all unsupported types from <a>observed entry types</a>. The user
```
yoavweiss: Do we define somewhere what are the supported types? Can we point to such a definition?
Also, can you rewrite that as an iteration over the sequence?

index.html
```diff
@@ -326,26 +331,35 @@ <h2><dfn>observe()</dfn> method</h2>
...
          <li>Otherwise, append a new <a>registered performance observer</a>
           object to the <a>list of registered performance observer objects</a> of
           <a>relevant global object</a>, with the <a>context object</a> as
           `observer` and <var>options</var> as the `options`.
           </li>
-          <li>If <var>options</var>' <a>buffered</a> flag is set, for each
-            <var>entry type</var> in <var>entry types</var> sequence:
+          <li>For each <var>entry type</var> in <a>observed entry types</a>, if the
+            <a data-link-for="PerformanceEntryObserverOptions">buffered</a> flag is set
+            in <var>options</var>' dictionary named <var>entry type</var>:
```
yoavweiss: `options` is a `PerformanceObserverInit`, right? Or is it `PerformanceEntryObserverOptions`?

index.html
```diff
@@ -366,14 +380,27 @@ <h2><dfn>PerformanceObserverInit</dfn> dictionary</h2>
...
           </pre>
           <dl>
             <dt><dfn>entryTypes</dfn></dt>
             <dd>A list of entry types to be observed. The list MUST NOT be empty
             and types not recognized by the user agent MUST be ignored.</dd>
           </dl>
+          <p class="note">Whenever a new kind of <a>PerformanceEntry</a> is defined,
+            the <a>PerformanceObserverInit</a> dictionary should be augmented to
+            include a new optional <a>PerformanceEntryObserverOptions</a> dictionary
+            with the name of the new <a data-lt="PerformanceEntry.entryType">
```
yoavweiss: It's not clear what you mean by "the name of the new entryType". Can you be more explicit?

index.html
npm1: I've added what observer() receives. I really do mean a link to the associated concept of the PerformanceObserver, not a local var. So <a> is correct in this context.

npm1: *so \<a\> is correct

index.html

index.html
npm1: What I mean is: add the entry types that are listed as dictionaries in 'options' to the set of entry types that are observed by this observer. In the example in the code, we had:
```
observer.observe({
  entryTypes: ["mark", "measure"],
  resource : {buffered: true}
});
```
In this case, there is a `resource` dictionary. So this step means: since the key `resource` is present, add 'resource' to the list of `observed entry types`.

I think @igrigorik wanted your opinion on whether this is sane or not. If we think this is too ugly, we can remove this part, so that all entry types to be observed must still be listed in the `entryTypes` array.

index.html
npm1: We do not define this (this is browser-dependent). If/when we expose `performance.entryTypes` then we can point to that, but for now I think we don't have a definition. Rewrote as an iteration (although note that this wording was already present before this PR).

index.html
npm1: Yes, but `buffered` would be inside the `PerformanceEntryObserverOptions` corresponding to that entry type (if it exists), which is itself inside the `PerformanceObserverInit` `options`. I moved this flag so that it can be an entryType-specific parameter.

index.html
npm1: What I mean is that `PerformanceObserverInit` needs to be augmented for example if we were adding EventTiming with entries that have entryType 'event':
```
partial dictionary PerformanceObserverInit {
   optional PerformanceEntryObserverOptions event;
}
```
I reworded that last part, is it better now?

rniwa:
I don't think this approach fits *Observer pattern well. See my comment in https://github.com/w3c/performance-timeline/issues/103#issuecomment-419296684.

npm1:
I'm open to changing the shape of the API but we probably want something that is backwards compatible. Changing observe() to only allow specifying one type per observe() call isn't.

rniwa:
We can support both. Either specify a list of types to observe, or specify a type and it's options.

npm1:
I'm closing this PR and later sending a new one based on the discussion in https://github.com/w3c/performance-timeline/issues/103

PR #108, opened 2018-10-05
Define supportedEntryTypes

npm1:
Add supportedEntryTypes and include an example of how it would be used. Solves https://github.com/w3c/performance-timeline/issues/77

***
<a href="https://pr-preview.s3.amazonaws.com/npm1/performance-timeline/pull/108.html" title="Last updated on Nov 26, 2018, 5:00 PM GMT (7e93eeb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/108/09c9d0a...npm1:7e93eeb.html" title="Last updated on Nov 26, 2018, 5:00 PM GMT (7e93eeb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -439,13 +439,13 @@
         buffer</a>.</p>
       </section>
       <section>
-        <h2><dfn>supportedEntryTypes</dfn> attribute</h2>
+        <h2><a>supportedEntryTypes</a> attribute</h2>
         <p>The user agent MUST maintain <dfn>supported entry types</dfn>, a list
         of strings representing the entry types which the user agent supports
         for the <a>PerformanceObserver</a> interface. This list is populated by
         specifications that define new entry types via the
-        <a href="#register-performance-entry-type"></a> algorithm.</p>
-        <p>When <a>supportedEntryTypes</a>'s attribute getter is called, run the following steps:</p>
+        <a>register a performance entry type</a> algorithm.</p>
+        <p>When <dfn>supportedEntryTypes</dfn>'s attribute getter is called, run the following steps:</p>
         <ol>
           <li>Let <var>result</var> be a copy of the user agent's <a>supported entry types</a>.</li>
           <li>Sort <var>result</var> in alphabetical order.</li>
@@ -571,7 +571,7 @@
     </section>
     <section data-link-for="PerformanceObserver">
       <h2>Register performance entry type</h2>
-      <p>To register a performance entry type, run the following steps:</p>
+      <p>To <dfn>register a performance entry type</dfn>, run the following steps:</p>
       <ol>
         <li>Let <var>type</var> be the input string.</li>
         <li>Append <var>type</var> to <a>supported entry types</a>.</li>
--- a/index.html
+++ b/index.html
@@ -353,7 +353,7 @@
           </li>
          <li>Otherwise, append a new <a>registered performance observer</a>
           object to the <a>list of registered performance observer objects</a> of
-          <a data-cite="!WHATWG-HTML#concept-relevant-global">relevant global object</a>, with the 
+          <a data-cite="!WHATWG-HTML#concept-relevant-global">relevant global object</a>, with the
           <a data-cite="!WHATWG-DOM#context-object">context object</a> as `observer` and <var>options</var> as the `options`.
           </li>
           <li>If <var>options</var>' <a>buffered</a> flag is set, for each
@@ -445,14 +445,17 @@
       </section>
       <section>
         <h2><dfn>supportedEntryTypes</dfn> attribute</h2>
-        <p>The user agent must maintain <dfn>supported entry types</dfn>, a list
+        <p>The user agent MUST maintain <dfn>supported entry types</dfn>, a list
         of strings representing the entry types which the user agent supports
         for the <a>PerformanceObserver</a> interface. This list is populated by
-        specs that define new entry types via the algorithm to
-        <a href="#register-performance-entry-type"></a>.</p>
-        <p>The <a>supportedEntryTypes</a>'s attribute getter must return a copy
-        of the user agent's <a>supported entry types</a>, sorted in alphabetical
-        order.</p>
+        specifications that define new entry types via the
+        <a href="#register-performance-entry-type"></a> algorithm.</p>
+        <p>When <a>supportedEntryTypes</a>'s attribute getter is called, run the following steps:</p>
+        <ol>
+          <li>Let <var>result</var> be a copy of the user agent's <a>supported entry types</a>.</li>
+          <li>Sort <var>result</var> in alphabetical order.</li>
+          <li>Return <var>result</var>.</li>
+        </ol>
         <p class="note">This attribute allows web developers to easily know
         which entry types are supported by the user agent.</p>
       </section>
--- a/index.html
+++ b/index.html
@@ -138,6 +138,15 @@
     &lt;body&gt;
     &lt;img id="image0" src="https://www.w3.org/Icons/w3c_main.png" /&gt;
     &lt;script&gt;
+    // Know when the entry types we would like to use are not supported.
+    function detectSupport(entryTypes) {
+      for (const entryType of entryTypes) {
+        if (!PerformanceObserver.supportedEntryTypes.includes(entryType)) {
+          console.log(entryType + " is not supported by PerformanceObserver!");
+        }
+      }
+    }
+    detectSupport(["resource", "mark", "measure"]);
     const observer = new PerformanceObserver(list =&gt; {
       list
         .getEntries()
@@ -151,12 +160,12 @@
           };
           return JSON.stringify(obj, null, 2);
         })
-        // Display them to the console
+        // Display them to the console.
         .forEach(console.log);
-      // maybe disconnect after processing the events.
+      // Maybe disconnect after processing the events.
       observer.disconnect();
     });
-    // retrieve buffered events and subscribe to new events
+    // Retrieve buffered events and subscribe to new events
     // for Resource-Timing and User-Timing
     observer.observe({
       entryTypes: ["resource", "mark", "measure"],
@@ -313,6 +322,7 @@
         void observe (PerformanceObserverInit options);
         void disconnect ();
         PerformanceEntryList takeRecords();
+        static readonly attribute FrozenArray&lt;DOMString&gt; supportedEntryTypes;
       };
       </pre>
       <p class="note">To keep the performance overhead to minimum the
@@ -429,6 +439,15 @@
         from the <a>list of registered performance observer objects</a> of
         <a>relevant global object</a>, and also empty <a>context object</a>'s
         <a>observer buffer</a>.</p>
+      </section>
+      <section>
+        <h2><dfn>supportedEntryTypes</dfn> attribute</h2>
+        <p>The <a>supportedEntryTypes</a>'s attribute getter must return a list
+        of entry types which the user agent supports for the
+        <a>PerformanceObserver</a> interface. This attribute allows web
+        developers to easily know which entry types are supported by the user
+        agent and to discover the entry types that exist and are supported by
+        the user agent.</p>
       </section>
     </section>
   </section>
--- a/index.html
+++ b/index.html
@@ -162,7 +162,7 @@
         })
         // Display them to the console.
         .forEach(console.log);
-      // Maybe disconnect after processing the events.
+      // Disconnect after processing the events.
       observer.disconnect();
     });
     // Retrieve buffered events and subscribe to new events
@@ -445,12 +445,17 @@
       </section>
       <section>
         <h2><dfn>supportedEntryTypes</dfn> attribute</h2>
-        <p>The <a>supportedEntryTypes</a>'s attribute getter must return a list
-        of entry types which the user agent supports for the
-        <a>PerformanceObserver</a> interface. This attribute allows web
-        developers to easily know which entry types are supported by the user
-        agent and to discover the entry types that exist and are supported by
-        the user agent.</p>
+        <p>The user agent must maintain <dfn>supported entry types</dfn>, a list
+        of strings representing the entry types which the user agent supports
+        for the <a>PerformanceObserver</a> interface. This list is populated by
+        specs that define new entry types via the hook to
+        <a href="#register-performance-entry-type"></a>.</p>
+        <p>The <a>supportedEntryTypes</a>'s attribute getter must return a copy
+        of the user agent's <a>supported entry types</a>, sorted in alphabetical
+        order.</p>
+        <p class="note">This attribute allows web developers to easily know
+        which entry types are supported by the user agent and to discover the
+        entry types that exist and are supported by the user agent.</p>
       </section>
     </section>
   </section>
@@ -551,6 +556,11 @@
         <li>Return the <a>list of entry objects</a>.
         </li>
       </ol>
+    </section>
+    <section data-link-for="PerformanceObserver">
+      <h2>Register performance entry type</h2>
+      <p>Given a string <var>type</var>, to register <var>type</var> the user
+        agent must append <var>type</var> to <a>supported entry types</a>.</p>
     </section>
   </section>
   <section id="privacy-security">
--- a/index.html
+++ b/index.html
@@ -118,7 +118,7 @@
     function detectSupport(entryTypes) {
       for (const entryType of entryTypes) {
         if (!PerformanceObserver.supportedEntryTypes.includes(entryType)) {
-          console.log(entryType + " is not supported by PerformanceObserver!");
+          // Indicate to client-side analytics that |entryType| is not supported.
         }
       }
     }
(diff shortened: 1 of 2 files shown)


npm1:
cc @spanicker

index.html
```diff
@@ -313,6 +322,7 @@ <h2>The <dfn>PerformanceObserver</dfn> interface</h2>
         void observe (PerformanceObserverInit options);
         void disconnect ();
         PerformanceEntryList takeRecords();
+        static readonly attribute FrozenArray&lt;DOMString&gt; supportedEntryTypes;
```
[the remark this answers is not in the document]
npm1: We could, I just thought this is a bit more clear. No strong opinions here.

index.html
```diff
@@ -430,6 +440,15 @@ <h2><dfn>disconnect()</dfn> method</h2>
         <a>relevant global object</a>, and also empty <a>context object</a>'s
         <a>observer buffer</a>.</p>
       </section>
+      <section>
+        <h2><dfn>supportedEntryTypes</dfn> attribute</h2>
```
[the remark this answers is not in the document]
npm1: Hmm I believe that's unnecessary right now: I doubt any web developer or browser would complain about the meaning of this being unclear. The benefit of registering to me would be unified behavior in terms of the buffers of each entry type. But that is currently not unified (i.e. our buffers have some differences depending on the entry type). So I'd see the value of the registration idea, but only if/when the buffers of all entry types behave the same way, so we could describe the entry types' buffer behavior in PerformanceTimeline. Does this make sense?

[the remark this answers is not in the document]
yoavweiss: IMO, adding this without such hooks would add more debt and under-defined concepts to the spec. As I commented below, "unsupported types" is already an under-defined concept, but I don't know if we want to add to its usage.

index.html
```diff
@@ -430,6 +440,15 @@ <h2><dfn>disconnect()</dfn> method</h2>
...
+        <p>The <a>supportedEntryTypes</a>'s attribute getter must return a list
```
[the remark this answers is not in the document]
npm1: I think it's clearer to use a static attribute here since the output is pretty fixed. I would use a static method when there is some computation involved (maybe with some needed parameters).

index.html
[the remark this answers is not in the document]
npm1: I disagree because in my view there is nothing to be defined about what "user agent support" means. It's also unclear when such hooks would be executed (it needs to be very early on to ensure this method never returns an incorrect value). So IMO the hooks could be added when there is some additional incentive to add them, not to define what supported entry types means. I do agree that we can improve the wording so that it is clear what we mean by these supported entry types, but I believe the hooks are not needed for this. To further argument my case with an example, I'd like to point out to this spec which exposes a similar attribute:
https://w3c.github.io/push-api/#dom-pushmanager-supportedcontentencodings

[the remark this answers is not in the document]
npm1: Oh ok I think I see some benefit of having a 'hook' now: if you're implementing say PaintTiming, you'd like to know that you need to update supportedEntryTypes accordingly. Thus, the PaintTiming spec should somehow mention the need for the user agent to update the array. Does this sound like the benefit that was being considered?

If that's the case, we can either a) explicitly call it out in each spec ('the user agent must not include "paint" in PerformanceObserver.supportedEntryTypes") or b) define a hook to "register entry type" which would do that. Ilya's initial sketch included buffer behavior, but like I said they currently behave differently so at this point we probably don't want to include buffers in the hook.

If any of those are resonable, what would be better, a) or b)? a) is simpler in the short term, b) means we can hook more shared implementation details later if needed.

index.html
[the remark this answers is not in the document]
yoavweiss: Fair enough :)

index.html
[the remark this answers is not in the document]
yoavweiss: > Oh ok I think I see some benefit of having a 'hook' now: if you're implementing say PaintTiming, you'd like to know that you need to update supportedEntryTypes accordingly. Thus, the PaintTiming spec should somehow mention the need for the user agent to update the array. Does this sound like the benefit that was being considered?

Exactly. We want to make sure the list remains up-to-date as new entry types are added, and the cleanest way to do that is to burden the dependent specs to update it.



> If that's the case, we can either a) explicitly call it out in each spec ('the user agent must not include "paint" in PerformanceObserver.supportedEntryTypes") or b) define a hook to "register entry type" which would do that. Ilya's initial sketch included buffer behavior, but like I said they currently behave differently so at this point we probably don't want to include buffers in the hook.

b) sounds like the better alternative, as it doesn't require the different dependent specs to be aware of each other. I agree we can start with simple entry type registration (the buffering related hooks are a different beast and we should tackle them separately).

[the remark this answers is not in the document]
rniwa: Yeah, I don't see how we can provide a hook for other specifications since we need to specify the order in which types are enumerated. We can't let multiple specifications specify the type without having a total order across them.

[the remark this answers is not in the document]
yoavweiss: Why do we need to specify order? If the order needs to be predictable, can't we sort the list in a predictable way?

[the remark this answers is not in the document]
rniwa: Oh, that's a good point. We should definitely sort the type names if we were to provide hooks.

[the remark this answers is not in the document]
igrigorik: As a thought experiment.. Instead of defining a "text hook" where upstream specs register types, could we instead just define this in IDL via [partial dictionary](https://heycam.github.io/webidl/#dfn-partial-dictionary) that each spec extends? Concretely..

1. PerfTimeline defines `entryTypes` which is an empty dictionary
1. Upstream specs provide partial's to extend said dictionary

Crazytalk?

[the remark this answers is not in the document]
rniwa: Why do we wan to use IDL for this? I don't think that makes much sense.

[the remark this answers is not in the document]
igrigorik: Registrations in prose are hard to track across specs and error prone due to nature of prose. With IDL, the registration is implicit — we don't need to define any hooks and such — and we can scan and pull in all the IDL declarations across specs, etc. Not saying we must do it this way, but worth considering.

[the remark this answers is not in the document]
npm1: I think the array I'm proposing is a bit cleaner. Dictionaries contain both keys and values, whereas here we would just need the keys (the strings). Also, WPT tests could help with making sure that the implementor updates `supportedEntryTypes` when needed.

index.html
```diff
@@ -430,6 +440,15 @@ <h2><dfn>disconnect()</dfn> method</h2>
...
+        of entry types which the user agent supports for the
```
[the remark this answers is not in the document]
npm1: I've added a hook

index.html
```diff
@@ -430,6 +440,15 @@ <h2><dfn>disconnect()</dfn> method</h2>
...
+        <a>PerformanceObserver</a> interface. This attribute allows web
+        developers to easily know which entry types are supported by the user
+        agent and to discover the entry types that exist and are supported by
+        the user agent.</p>
```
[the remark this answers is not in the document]

npm1:
BTW this is ready for review again, please take a look. Thanks!

index.html
```diff
@@ -433,6 +443,20 @@ <h2><dfn>disconnect()</dfn> method</h2>
...
+        of strings representing the entry types which the user agent supports
+        for the <a>PerformanceObserver</a> interface. This list is populated by
+        specs that define new entry types via the hook to
+        <a href="#register-performance-entry-type"></a>.</p>
+        <p>The <a>supportedEntryTypes</a>'s attribute getter must return a copy
+        of the user agent's <a>supported entry types</a>, sorted in alphabetical
+        order.</p>
+        <p class="note">This attribute allows web developers to easily know
+        which entry types are supported by the user agent and to discover the
+        entry types that exist and are supported by the user agent.</p>
```
igrigorik: Seems like we're repeating ourselves three times here? Could we simplify to..

> This attribute allows web developers to easily know which entry types are supported by the user agent.

index.html
```diff
@@ -433,6 +443,20 @@ <h2><dfn>disconnect()</dfn> method</h2>
         <a data-cite="!WHATWG-DOM#context-object">context object</a>'s
         <a>observer buffer</a>.</p>
       </section>
+      <section>
+        <h2><dfn>supportedEntryTypes</dfn> attribute</h2>
+        <p>The user agent must maintain <dfn>supported entry types</dfn>, a list
+        of strings representing the entry types which the user agent supports
+        for the <a>PerformanceObserver</a> interface. This list is populated by
```
igrigorik: This list is initially empty and extended by other specifications by calling the ... algorithm?

_p.s. this also means we need to go back and update all the specs that emit such events._

index.html
```diff
@@ -533,6 +557,11 @@ <h2>Filter <var>buffer</var> by <var>name</var> and <var>type</var></h2>
         </li>
       </ol>
     </section>
+    <section data-link-for="PerformanceObserver">
+      <h2>Register performance entry type</h2>
+      <p>Given a string <var>type</var>, to register <var>type</var> the user
+        agent must append <var>type</var> to <a>supported entry types</a>.</p>
```
igrigorik: s/must/MUST/ since it's a special keyword?

index.html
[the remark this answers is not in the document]
yoavweiss: Personally, I prefer the ` supported` prefix, as it makes it clearer what information the attribute gives us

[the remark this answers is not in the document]
igrigorik: Fair enough. Ship it.

index.html

index.html
npm1: Done, and ack

rniwa:
Hm... [`DOMTokenList`](https://dom.spec.whatwg.org/#interface-domtokenlist) has `supports` method as opposed to a list of tokens supported. Perhaps matching that pattern is better. This will also avoid the ordering issue.

npm1:
Consistency with completely unrelated APIs is probably not something we should worry too much about. Otherwise I could say we should also be consistent with [PushManager](https://w3c.github.io/push-api/#dom-pushmanager-supportedcontentencodings) and we will never be happy :) A better question is what would be more ergonomic to web developers. I think both ways are valid so don't have a strong opinion on which is better.

rniwa:
You say that but all these inconsistencies throughout various features is the worst thing about the Web platform. Ideally, TAG or some other group familiar with all aspects of the Web API should be noticing these differences and fixing them but that doesn't seem to be happening either.

igrigorik:
@rniwa hmm, I like the idea — good catch. One gotcha I see here is that it also supports add/remove/toggle, which are a bit odd in our context. That said, if we're returning an array, then perhaps there is no difference?

index.html
```diff
@@ -138,6 +138,15 @@ <h2>Introduction</h2>
...
     &lt;script&gt;
+    // Know when the entry types we would like to use are not supported.
+    function detectSupport(entryTypes) {
+      for (const entryType of entryTypes) {
+        if (!PerformanceObserver.supportedEntryTypes.includes(entryType)) {
+          console.log(entryType + " is not supported by PerformanceObserver!");
+        }
+      }
+    }
+    detectSupport(["resource", "mark", "measure"]);
```
yoavweiss: This example is correct but not necessarily very useful. Maybe show an example of "if supported, register a PO, otherwise do something else (e.g. poll the timeline)"

index.html
```diff
@@ -433,6 +443,19 @@ <h2><dfn>disconnect()</dfn> method</h2>
         <a data-cite="!WHATWG-DOM#context-object">context object</a>'s
         <a>observer buffer</a>.</p>
       </section>
+      <section>
+        <h2><dfn>supportedEntryTypes</dfn> attribute</h2>
+        <p>The user agent must maintain <dfn>supported entry types</dfn>, a list
+        of strings representing the entry types which the user agent supports
+        for the <a>PerformanceObserver</a> interface. This list is populated by
+        specs that define new entry types via the algorithm to
```
yoavweiss: Nits:
s/specs/specifications/
s/via the algorithm to register performance entry type/via the register performance entry type algorithm/

index.html
```diff
@@ -533,6 +556,11 @@ <h2>Filter <var>buffer</var> by <var>name</var> and <var>type</var></h2>
         </li>
       </ol>
     </section>
+    <section data-link-for="PerformanceObserver">
+      <h2>Register performance entry type</h2>
+      <p>Given a string <var>type</var>, to register <var>type</var> the user
+        agent MUST append <var>type</var> to <a>supported entry types</a>.</p>
```
yoavweiss: Can we define a simple algorithm here, instead of these "MUST" semantics?
e.g. "1. Let `type` be the input string to the algorithm 2. Append `type` to `supported entry types`"

index.html
```diff
@@ -433,6 +443,19 @@ <h2><dfn>disconnect()</dfn> method</h2>
         <a data-cite="!WHATWG-DOM#context-object">context object</a>'s
         <a>observer buffer</a>.</p>
       </section>
+      <section>
+        <h2><dfn>supportedEntryTypes</dfn> attribute</h2>
+        <p>The user agent must maintain <dfn>supported entry types</dfn>, a list
```
yoavweiss: s/must/MUST/

index.html
```diff
@@ -433,6 +443,19 @@ <h2><dfn>disconnect()</dfn> method</h2>
...
+        of strings representing the entry types which the user agent supports
+        for the <a>PerformanceObserver</a> interface. This list is populated by
+        specs that define new entry types via the algorithm to
+        <a href="#register-performance-entry-type"></a>.</p>
+        <p>The <a>supportedEntryTypes</a>'s attribute getter must return a copy
```
yoavweiss: Here too, may be better to convert that into an algorithm and get rid of the "must". (e.g. 
"when the attribute getter is called, run the following steps: ...")

yoavweiss:
> Hm... [`DOMTokenList`](https://dom.spec.whatwg.org/#interface-domtokenlist) has `supports` method as opposed to a list of tokens supported. Perhaps matching that pattern is better. This will also avoid the ordering issue.

`supports` was added to `DOMTokenList` as a way to feature detect supported attribute values. This is not quite the same case. We already have a way to [feature detect](https://github.com/w3c/performance-timeline/issues/64#issuecomment-273758093) support for the various features. This will just provide a cleaner way to do the same as well as discover observable types.

index.html
npm1: Why would we poll the timeline when unsupported? If an entry type is unsupported in PerformanceObserver then it won't be buffered.

index.html
npm1: I don't see the benefit here, it would be a one-step (two if you count mentioning the input as one step) algorithm with no return value... And also this is supposed to be implicitly done by the user agent when implementing a specification, not a normal algorithm that can be called at any time.

index.html
yoavweiss: On second thought, this is not supposed to resolve feature detection necessarily, so my example doesn't necessarily make sense, at least not in practice. Maybe instead of (or alongside) a console log, you can also turn on a flag that indicates the client-side analytics code that there's no need to way for unsupported PO callbacks.

index.html
yoavweiss: It's cleaner. Right now IIUC you want to have other specifications link to the section itself. It's cleaner to have them link to an algorithm definition, and for that algorithm to actually be well-defined.
I agree that it is super simple and doesn't return a value, but it'd still be cleaner.


> And also this is supposed to be implicitly done by the user agent when implementing a specification, not a normal algorithm that can be called at any time.

It is supposed to be called from different specifications that implement PO.

index.html
npm1: Ok, I've replaced the console.log with a comment.

index.html
npm1: Ok I wrote it as a 2-step algorithm.

index.html
```diff
@@ -543,6 +569,14 @@ <h2>Filter <var>buffer</var> by <var>name</var> and <var>type</var></h2>
         </li>
       </ol>
     </section>
+    <section data-link-for="PerformanceObserver">
+      <h2>Register performance entry type</h2>
+      <p>To register a performance entry type, run the following steps:</p>
```
yoavweiss: you should put "register a performance entry type" in a `<dfn>`, so that other specs can hook into it

index.html
```diff
@@ -428,6 +438,22 @@ <h2><dfn>disconnect()</dfn> method</h2>
         "WHATWG-DOM#context-object">context object</a>'s <a>observer
         buffer</a>.</p>
       </section>
+      <section>
+        <h2><dfn>supportedEntryTypes</dfn> attribute</h2>
```
yoavweiss: I don't think 2 `<dfn>` to the same attribute, so this should probably be an `<a>` (with the dfn being in the paragraph below)

index.html
yoavweiss: Once you add that, you should also link directly to the algorithm in line 447, rather than to the section

index.html
npm1: Changed the dfn to the place where attribute getter is described. However, there's not quite repetition. There's `supportedEntryTypes`, the attribute, and `supported entry types`, the list maintained by the user agent which is used as the return value.

npm1:
I'll followup with modifications to other specs as well as tests.

npm1:
Yoav, there should not be a need for such a test. That's what idlharness is for. When the spec crawler gets updated, it should start failing for implementations that do not have this attribute implemented. So, are you ok just landing this?

yoavweiss:
Makes sense. Ship it! :)

PR #112, opened 2018-12-07
Add |type| to PerformanceObserverInit

npm1:
This change introduces a DOMString type in PerformanceObserverInit to support observing a single entryType with other parameters from the dictionary. The change modifies the observe() method so that:
* observe({entryTypes:...}) is still supported but not with any other member of the PerformanceObserverInit dictionary.
* observe({type: ...}) is now supported, and multiple calls stack.
* Both ways to call observe() cannot be mixed. That is, an observer has to stick to one of the ways, and 'observer type' is introduced to enforce this.

In addition, some nits are fixed along the way, such as using <var> and <a> appropriately throughout.
Solves https://github.com/w3c/performance-timeline/issues/103

***
<a href="https://pr-preview.s3.amazonaws.com/npm1/performance-timeline/pull/112.html" title="Last updated on Dec 14, 2018, 10:01 PM UTC (f86cf35)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/112/520de3e...npm1:f86cf35.html" title="Last updated on Dec 14, 2018, 10:01 PM UTC (f86cf35)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -330,7 +330,7 @@
       };
       </pre>
       <p class="note">To keep the performance overhead to minimum the
-      application should only subscribe to event types that it is interested
+      application ought to only subscribe to event types that it is interested
       in, and disconnect the observer once it no longer needs to observe the
       performance data. Filtering by name is not supported, as it would
       implicitly require a subscription for all event types — this is possible,
@@ -454,7 +454,7 @@
           </ol>
           </li>
         </ol>
-        <p class="note">A <a>PerformanceObserver</a> object should always call
+        <p class="note">A <a>PerformanceObserver</a> object needs to always call
         <a>observe()</a> with <var>options</var>'s
         <a data-link-for=PerformanceObserverInit>entryTypes</a> set OR always
         call <a>observe()</a> with <var>options</var>'s
--- a/index.html
+++ b/index.html
@@ -162,7 +162,7 @@
       // Disconnect after processing the events.
       resourceObserver.disconnect();
     });
-    // Retrieve buffered events and subscribe to never events for Resource Timing.
+    // Retrieve buffered events and subscribe to newer events for Resource Timing.
     resourceObserver.observe({type: "resource", buffered: true});
     &lt;/script&gt;
     &lt;/body&gt;
@@ -338,15 +338,17 @@
       <section>
         <h2><dfn>observe()</dfn> method</h2>
         <p>The <a>observe()</a> method instructs the user agent to
-        <dfn>register the <var>observer</var></dfn> and must run these steps:</p>
+        <dfn>register the observer</dfn> and must run these steps:</p>
         <ol data-link-for="PerformanceObserverInit">
+          <li>Let <var>observer</var> be the <a data-cite=
+            "WHATWG-DOM#context-object">context object</a>.</li>
           <li>If <var>options</var>'s <a>entryTypes</a> and <a>type</a> members
-          are both omitted, then <a data-cite="!WEBIDL#dfn-throw">throw</a> a
-          <a data-cite="!WEBIDL#syntaxerror"><code>SyntaxError</code></a>.</li>
+          are both omitted, then <a data-cite="WEBIDL#dfn-throw">throw</a> a
+          <a data-cite="WEBIDL#syntaxerror"><code>SyntaxError</code></a>.</li>
           <li>If <var>options</var>'s <a>entryTypes</a> is present and any other
           member is also present, then
-          <a data-cite="!WEBIDL#dfn-throw">throw</a> a
-          <a data-cite="!WEBIDL#syntaxerror"><code>SyntaxError</code></a>.</li>
+          <a data-cite="WEBIDL#dfn-throw">throw</a> a
+          <a data-cite="WEBIDL#syntaxerror"><code>SyntaxError</code></a>.</li>
           <li>Update or check <var>observer</var>'s <a>observer type</a> by
           running these steps:
             <ol>
@@ -364,14 +366,14 @@
               <li>If <var>observer</var>'s <a>observer type</a> is
               <code>"single"</code> and <var>options</var>'s
               <a>entryTypes</a> member is present, then
-              <a data-cite="!WEBIDL#dfn-throw">throw</a> an
-              <a data-cite="!WEBIDL#invalidmodificationerror">
+              <a data-cite="WEBIDL#dfn-throw">throw</a> an
+              <a data-cite="WEBIDL#invalidmodificationerror">
               <code>InvalidModificationError</code></a>.
               </li>
               <li>If <var>observer</var>'s <a>observer type</a> is
               <code>"multiple"</code> and <var>options</var>'s <a>type</a>
-              member is present, then <a data-cite="!WEBIDL#dfn-throw">throw</a>
-              an <a data-cite="!WEBIDL#invalidmodificationerror">
+              member is present, then <a data-cite="WEBIDL#dfn-throw">throw</a>
+              an <a data-cite="WEBIDL#invalidmodificationerror">
               <code>InvalidModificationError</code></a>.
               </li>
             </ol>
@@ -398,9 +400,9 @@
             "WHATWG-DOM#context-object">context object</a>, replace its
             <a>options list</a> with a list containing <var>options</var> as its
             only item.</li>
-            <li>Otherwise, append a new <a>registered performance observer</a>
-            object to the <a>list of registered performance observer objects</a>
-            of <a data-cite=
+            <li>Otherwise, create and append a <a>registered performance
+            observer</a> object to the <a>list of registered performance
+            observer objects</a> of <a data-cite=
             "HTML/webappapis.html#concept-relevant-global">relevant global
             object</a>, with <a>observer</a> set to the
             <a data-cite="WHATWG-DOM#context-object">context object</a> and
@@ -408,9 +410,10 @@
             its only item.</li>
           </ol>
           </li>
-          <li>Otherwise (<var>observer</var>'s <a>observer type</a> is
-          <code>"single"</code>), run the following steps:
+          <li>Otherwise, run the following steps:
           <ol>
+            <li>Assert that <var>observer</var>'s <a>observer type</a> is
+            <code>"single"</code>.</li>
             <li>If <var>options</var>'s <a>type</a> is not contained in
             <a>supported entry types</a>, abort these steps. The user agent
             SHOULD notify developers when this happens, for instance via a
@@ -426,16 +429,18 @@
                 whose <a>type</a> is equal to <var>options</var>'s <a>type</a>,
                 replace <var>currentOptions</var> with <var>options</var> in
                 <var>obs</var>'s <a>options list</a>.</li>
-                <li>Otherwise, append a new
-                <a>registered performance observer</a> object to the <a>list of
-                registered performance observer objects</a> of
-                <a data-cite="HTML/webappapis.html#concept-relevant-global">
-                relevant global object</a>, with <a>observer</a> set to the
-                <a data-cite="WHATWG-DOM#context-object">context object</a> and
-                <a>options list</a> set to a list containing <var>options</var>
-                as its only item.</li>
+                <li>Otherwise, append <var>options</var> to <var>obs</var>'s
+                <a>options list</a>.</li>
               </ol>
             </li>
+            <li>Otherwise, create and append a
+            <a>registered performance observer</a> object to the <a>list of
+            registered performance observer objects</a> of
+            <a data-cite="HTML/webappapis.html#concept-relevant-global">
+            relevant global object</a>, with <a>observer</a> set to the
+            <a data-cite="WHATWG-DOM#context-object">context object</a> and
+            <a>options list</a> set to a list containing <var>options</var>
+            as its only item.</li>
             <li>If <var>options</var>'s <a>buffered</a> flag is set:
               <ol>
                 <li>Let <var>entries</var> be the <a>PerformanceEntryList</a>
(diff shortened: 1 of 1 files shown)


npm1:
@yoavweiss @igrigorik @rniwa @toddreifsteck please take a look, this is ready for review.

index.html
```diff
@@ -319,64 +338,155 @@ <h2>The <dfn>PerformanceObserver</dfn> interface</h2>
...
-          of <a data-cite=
-          "HTML/webappapis.html#concept-relevant-global">relevant global
-          object</a>, with the <a data-cite="WHATWG-DOM#context-object">context
-          object</a> as `observer` and <var>options</var> as the `options`.
-          </li>
-          <li>If <var>options</var>' <a>buffered</a> flag is set, for each
-          <var>entry type</var> in <var>entry types</var> sequence:
+          <li>If <var>options</var>'s <a>entryTypes</a> and <a>type</a> members
+          are both omitted, then <a data-cite="!WEBIDL#dfn-throw">throw</a> a
+          <a data-cite="!WEBIDL#syntaxerror"><code>SyntaxError</code></a>.</li>
```
igrigorik: Sanity check: is SyntaxError the best fit here?

index.html
```diff
@@ -319,64 +338,155 @@ <h2>The <dfn>PerformanceObserver</dfn> interface</h2>
...
+                <li>Append <var>entries</var> to the <a data-cite=
+                "WHATWG-DOM#context-object">context object</a>'s <a>observer
+                buffer</a>.
+                </li>
+              </ol>
+            </li>
+          </ol>
+          </li>
         </ol>
+        <p class="note">A <a>PerformanceObserver</a> object should always call
```
igrigorik: fwiw, since this is non-normative text, we might want to use alternates for should, must, etc: https://tools.ietf.org/id/draft-hansen-nonkeywords-non2119-04.html#rfc.section.1

index.html
npm1: Hmm I'm using it because this is the type of error used in https://w3c.github.io/user-timing/#mark-method and it seems similar. Any thoughts on what type of error could be better?

index.html
npm1: Done and fixed the other note which also had a should

npm1:
Other folks - thoughts?

index.html
```diff
@@ -139,14 +139,31 @@ <h2>Introduction</h2>
...
+            "Response End": responseEnd,
+          };
+          return JSON.stringify(obj, null, 2);
+        })
+        // Display them to the console.
+        .forEach(console.log);
+      // Disconnect after processing the events.
+      resourceObserver.disconnect();
     });
+    // Retrieve buffered events and subscribe to never events for Resource Timing.
```
yoavweiss: s/never/newer/

index.html
igrigorik: Fair enough. I don't have a competing alternative, just wanted to flag for other reviewers in case there is a better option here.

igrigorik:
@rniwa @toddreifsteck would love to hear your input on this one.

npm1:
Maybe directly add them as reviewers in case they have email filters for mentions :) I can't add reviewers here.

index.html
```diff
@@ -311,72 +330,163 @@ <h2>The <dfn>PerformanceObserver</dfn> interface</h2>
...
+      application ought to only subscribe to event types that it is interested
       in, and disconnect the observer once it no longer needs to observe the
       performance data. Filtering by name is not supported, as it would
       implicitly require a subscription for all event types — this is possible,
       but discouraged, as it will generate a significant volume of events.</p>
       <section>
         <h2><dfn>observe()</dfn> method</h2>
         <p>The <a>observe()</a> method instructs the user agent to
-        <dfn>register the observer</dfn> and must run these steps:</p>
+        <dfn>register the <var>observer</var></dfn> and must run these steps:</p>
```
yoavweiss: For both observer and options, I prefer "Let observer be..." style where the variables are explicitly defined

index.html
```diff
@@ -311,72 +330,163 @@ <h2>The <dfn>PerformanceObserver</dfn> interface</h2>
...
-              "WHATWG-DOM#context-object">context object</a>'s <a>observer
-              buffer</a>.
+              <li>If <var>observer</var>'s <a>observer type</a> is
+              <code>"single"</code> and <var>options</var>'s
+              <a>entryTypes</a> member is present, then
+              <a data-cite="!WEBIDL#dfn-throw">throw</a> an
+              <a data-cite="!WEBIDL#invalidmodificationerror">
+              <code>InvalidModificationError</code></a>.
+              </li>
+              <li>If <var>observer</var>'s <a>observer type</a> is
```
yoavweiss: Similarly, what's the case in which we reach here?

index.html
yoavweiss: Also, observer here is used both as a var and as a member of a potential PO. Can you pick a different name for the local variable?

index.html
```diff
@@ -311,72 +330,163 @@ <h2>The <dfn>PerformanceObserver</dfn> interface</h2>
...
+            when the steps are aborted to notify that registration has been
+            aborted. For example, a console warning might be appropriate.</li>
+            <li>If the <a>list of registered performance observer objects</a> of
+            <a data-cite="HTML/webappapis.html#concept-relevant-global">relevant
+            global object</a> contains a <a>registered performance observer</a>
+            whose <a>observer</a> is the <a data-cite=
+            "WHATWG-DOM#context-object">context object</a>, replace its
+            <a>options list</a> with a list containing <var>options</var> as its
+            only item.</li>
+            <li>Otherwise, append a new <a>registered performance observer</a>
```
yoavweiss: create and append?

index.html
```diff
@@ -311,72 +330,163 @@ <h2>The <dfn>PerformanceObserver</dfn> interface</h2>
...
+            object to the <a>list of registered performance observer objects</a>
+            of <a data-cite=
+            "HTML/webappapis.html#concept-relevant-global">relevant global
+            object</a>, with <a>observer</a> set to the
+            <a data-cite="WHATWG-DOM#context-object">context object</a> and
+            <a>options list</a> set to a list containing <var>options</var> as
+            its only item.</li>
+          </ol>
+          </li>
+          <li>Otherwise (<var>observer</var>'s <a>observer type</a> is
```
yoavweiss: Maybe instead of the comment, assert that observer type is single?

index.html
```diff
@@ -311,72 +330,163 @@ <h2>The <dfn>PerformanceObserver</dfn> interface</h2>
...
+        <a data-link-for=PerformanceObserverInit>entryTypes</a> set OR always
+        call <a>observe()</a> with <var>options</var>'s
+        <a data-link-for=PerformanceObserverInit>type</a> set. This is meant to
+        avoid confusion with how calls would stack. When using
+        <a data-link-for=PerformanceObserverInit>entryTypes</a>, no other
+        parameters in <a>PerformanceObserverInit</a> can be used. In addition,
+        multiple <a>observe()</a> calls will override for backwards
+        compatibility and because a single call should suffice in this case. On
+        the other hand, when using
+        <a data-link-for=PerformanceObserverInit>type</a>, calls will stack
```
yoavweiss: You also want to add a note that calling `observe()` with entry types and then `observe()` with type (and vice versa) will throw

yoavweiss: Also, when calling observe with the same type multiple times, it will override, right?

index.html
```diff
@@ -311,72 +330,163 @@ <h2>The <dfn>PerformanceObserver</dfn> interface</h2>
...
-            <dd>A list of entry types to be observed. The list MUST NOT be
-            empty and types not recognized by the user agent MUST be
-            ignored.</dd>
+            <dd>A list of entry types to be observed. If present, the list MUST
+            NOT be empty and all other members MUST NOT be present. Types not
+            recognized by the user agent MUST be ignored.</dd>
+          </dl>
+          <dl>
+            <dt><dfn>type</dfn></dt>
+            <dd>A single entry type to be observed. If present, the type MUST be
```
yoavweiss: What do you mean by "MUST be recognized" here? If the type is unknown, the UA MUST ignore it, no?

index.html
```diff
@@ -464,27 +574,32 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
...
-            <li>If <i>observer</i>'s <a>PerformanceObserverInit</a>
-              <a data-lt="PerformanceObserverInit.entryTypes">entryTypes</a>
-              includes <i>new entry</i>’s <a data-lt=
-              "PerformanceEntry.entryType">entryType</a> value, append
-              <i>observer</i> to <i>interested observers</i>.
+            <li>If <var>regObs</var>'s <a>options list</a> contains a
+            <a>PerformanceObserverInit</a> item whose
+            <a data-link-for="PerformanceObserverInit">entryTypes</a> member
+            includes <var>entryType</var> or whose
+            <a data-link-for="PerformanceObserverInit">type</a> member is
```
yoavweiss: s/is/equals to/

index.html
```diff
@@ -542,30 +657,30 @@ <h2>Filter <var>buffer</var> by <var>name</var> and <var>type</var></h2>
       respect to <a>startTime</a>; objects with the same <a>startTime</a> have
       unspecified ordering.</p>
       <ol>
-        <li>Let the <dfn>list of entry objects</dfn> be the empty
+        <li>Let the <var>list of entry objects</var> be the empty
```
yoavweiss: Paging Dr. @marcoscaceres for opinions on `<i>` vs. `<var>` here

index.html
```diff
@@ -502,21 +617,21 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
...
-                <li>Let <i>entries</i> be a copy of <i>po</i>’s <a>observer
+                <li>Let <var>entries</var> be a copy of <var>po</var>’s <a>observer
                 buffer</a>.
                 </li>
-                <li>Empty <i>po</i>’s <a>observer buffer</a>.
+                <li>Empty <var>po</var>’s <a>observer buffer</a>.
                 </li>
-                <li>If <i>entries</i> is non-empty, call <i>po</i>’s callback
-                with <i>entries</i> as first argument and <i>po</i> as the
+                <li>If <var>entries</var> is non-empty, call <var>po</var>’s callback
```
yoavweiss: Paging Dr. @marcoscaceres for opinions on `<i>` vs. `<var>` here

index.html

index.html

index.html
```diff
@@ -311,72 +330,163 @@ <h2>The <dfn>PerformanceObserver</dfn> interface</h2>
...
-          of <a data-cite=
-          "HTML/webappapis.html#concept-relevant-global">relevant global
-          object</a>, with the <a data-cite="WHATWG-DOM#context-object">context
-          object</a> as `observer` and <var>options</var> as the `options`.
-          </li>
-          <li>If <var>options</var>' <a>buffered</a> flag is set, for each
-          <var>entry type</var> in <var>entry types</var> sequence:
+          <li>If <var>options</var>'s <a>entryTypes</a> and <a>type</a> members
+          are both omitted, then <a data-cite="!WEBIDL#dfn-throw">throw</a> a
+          <a data-cite="!WEBIDL#syntaxerror"><code>SyntaxError</code></a>.</li>
```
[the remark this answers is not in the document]

index.html
npm1: Hm sure, but then probably remove the 'new' since it would be redundant?

index.html

index.html

index.html

npm1:
Addressed feedback, shall continue the conversation next year! :)

index.html
marcoscaceres: As a rule, avoid `<i>` for variables. Always use `<var>` instead of `<i>`, unless you really mean "italics" (which would be weird?). ReSpec treats `var`s as special, allow readers to click on them to see where they are used. Additionally, `<var>`s will soon know about their IDL or Infra data type, making them more powerful.

npm1: Ok, so `var` seems correct here.

npm1:
Hey @yoavweiss could you take another look?

index.html
```diff
@@ -321,62 +340,164 @@ <h2><dfn>observe()</dfn> method</h2>
...
-                type</var>.
+              <li>If <var>observer</var>'s <a>observer type</a> is
+              <code>"undefined"</code>:
+                <ol>
+                  <li>If <var>options</var>'s <a>entryTypes</a> member is
+                  present, then set <var>observer</var>'s <a>observer type</a>
+                  to <code>"multiple"</code>.</li>
+                  <li>If <var>options</var>'s <a>type</a> member is present,
+                  then set <var>observer</var>'s <a>observer type</a> to
+                  <code>"single"</code>.</li>
```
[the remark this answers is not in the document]
npm1: We throw. That is in step 2.

index.html
```diff
@@ -321,62 +340,164 @@ <h2><dfn>observe()</dfn> method</h2>
...
+        <a>observe()</a> with <var>options</var>'s
+        <a data-link-for=PerformanceObserverInit>entryTypes</a> set OR always
+        call <a>observe()</a> with <var>options</var>'s
+        <a data-link-for=PerformanceObserverInit>type</a> set. If one
+        <a>PerformanceObserver</a> calls <a>observe()</a> with <a data-link-for=
+        PerformanceObserverInit>entryTypes</a> and also calls observe with
+        <a data-link-for=PerformanceObserverInit>type</a>, then an exception is
+        thrown. This is meant to avoid confusion with how calls would stack.
+        When using <a data-link-for=PerformanceObserverInit>entryTypes</a>, no
+        other parameters in <a>PerformanceObserverInit</a> can be used. In
```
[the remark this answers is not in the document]
npm1: They won't be ignored. I throw an exception if there are other parameters (step 3)

index.html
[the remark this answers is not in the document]
yoavweiss: OK, cool

npm1:
Yes I'll add the tests alongside Chrome implementation, which will followup soon, if that's ok.

yoavweiss:
> Yes I'll add the tests alongside Chrome implementation, which will followup soon, if that's ok.

SGTM

index.html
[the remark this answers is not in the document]
yoavweiss: That's actually problematic from a future compatibility perspective. If we'd want to add more parameters in the future, that would make it more difficult.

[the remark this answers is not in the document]
npm1: The idea is that if you want to use parameters then you should use `type`, not `entryTypes`. This enforces that behavior as opposed to silently letting it pass

Issue #117, opened 2019-03-27
supportedEntryTypes API has some issues


bzbarsky:
The two (interrelated) problems are:

1) It's an attribute that is returning a new value each time.  This is a bad enough antipattern, known for long enough, that Web IDL tries to prevent it; it looks like the definition of this API used `FrozenArray` instead of sequence to try to work around that prevention.

2) The workaround is broken: the steps for the attribute getter return an infra list, but per Web IDL if you are returning a value of type `FrozenArray` you need to return an actual frozen Array object.  There is no implicit conversion from _list_ defined at https://heycam.github.io/webidl/#idl-frozen-array or https://heycam.github.io/webidl/#es-frozen-array or https://infra.spec.whatwg.org/#lists (compare to sequence, which _does_ define such a conversion at https://heycam.github.io/webidl/#idl-sequence

If we want to return a new object each time here, this should be a method, not an attribute.  If we want to return the same object each time, then using a FrozenArray makes sense, but it's worth thinking about where that object is going to be cached, because it needs to be cached somewhere.  Probably in a slot on either the PerformanceObserver interface object or the global.

My personal preference would be for this API to be a static method returning a sequence, fwiw.

@npm1 @domenic @foolip @yoavweiss @igrigorik @spanicker @bakulf @w3c/web-assembly 

P.S. I am aware that Chrome has already shipped this, with the "return a new value each time" behavior.  I am fairly opposed to Gecko shipping anything of the sort.

npm1:
Hi, it should be an attribute that returns the same object each time. Will send a PR soon to fix, and will fix Chrome implementation. Feel free to implement in Gecko without returning a new value each time. Regarding where it should be cached, I think it's related to https://github.com/w3c/performance-timeline/issues/113. The list of supported entry types may depend on the execution context, so it should be cached per execution context.

bzbarsky:
The caching matters because it affects which object is returned, observably.

So in particular, if you take the getter from one global and call it with a `this` value that is `PerformanceObserver` interface object from another global, which global does the value come from?

For that matter, what happens if you call it with a `this` value that is not a `PerformanceObserver` object at all?  https://heycam.github.io/webidl/#dfn-attribute-getter step 1.1.2 does nothing for static attributes, so they don't enforce any constraints on their "this" value.

Presumably you can tie the value to the current global of the getter.  It implies a bit of weirdness if someone moves the function around across globals, but that might be ok.

bzbarsky:
[...]

Issue #113, opened 2019-01-16
supported entry types needs to be context aware


cvazac:
Currently, [`supported entry types`](https://github.com/w3c/performance-timeline/blob/gh-pages/index.html#L564) is specific to the user agent. It also needs to be _context_ (window v. worker) aware, because `navigation` entries are **not** observable from workers, for example.

toddreifsteck:
Triaged this for possible fix in L2. Multiple other specifications most likely need their registrations to be context aware and ideally should build on language in this spec. They can monkey patch if needed but it is not ideal.

npm1:
@cvazac assigning to you per the discussion during the call

PR #118, opened 2019-03-27
supportedEntryTypes: cache the FrozenArray

npm1:
This change does the following:
* Cache the result from PerformanceObserver's attribute getter in |frozen supported entry types|, and mention that other specs determine its contents.
* Remove the non-frozen list and the hook to add to that list.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/pull/118.html" title="Last updated on May 6, 2019, 9:39 PM UTC (08f743c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/118/58b1f72...08f743c.html" title="Last updated on May 6, 2019, 9:39 PM UTC (08f743c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -199,23 +199,22 @@
       <li>A <dfn>performance entry buffer</dfn> to store
       <a>PerformanceEntry</a> objects that is initially empty.
       </li>
-      <li>A list of <dfn>supported entry types</dfn>, which contains the strings
-        representing the entry types which the user agent supports for the
-        <a>PerformanceObserver</a> interface within that environment. This list
-        is populated by specifications that define new entry types via the
-        <a>register a performance entry type</a> algorithm.
-      </li>
       <li>
         A <dfn>frozen supported entry types</dfn> representing the cached
         <a data-cite="WebIDL#idl-frozen-array">FrozenArray</a> for that
-        <a data-cite="WebIDL#es-environment">ECMAScript global environment</a>,
-        which is initially <code>null</code>.
+        <a data-cite="WebIDL#es-environment">ECMAScript global environment</a>.
+        The contents of this array are determined by specifications which define
+        new entry types.
       </li>
     </ul>
     <p>The <dfn>relevant performance entry buffer</dfn> is the <a>performance
     entry buffer</a> associated with the <a data-cite=
     "HTML/webappapis.html#concept-relevant-global">relevant global
     object</a>.</p>
+    <p>The <dfn>relevant supported entry types</dfn> is the <a>frozen supported
+    entry types</a> associated with the <a data-cite=
+    "HTML/webappapis.html#concept-relevant-global">relevant global object</a>.
+    </p>
     <section data-dfn-for="Performance" data-link-for="Performance">
       <h2>Extensions to the <dfn data-cite=
       "hr-time-2#sec-performance">Performance</dfn> interface</h2>
@@ -397,10 +396,10 @@
             <a>entryTypes</a> sequence.
             </li>
             <li>Remove all types from <var>entry types</var> that are not
-            contained in <a>supported entry types</a>. The user agent SHOULD
-            notify developers if <var>entry types</var> is modified. For
-            example, a console warning listing removed types might be
-            appropriate.</li>
+            contained in the <a>relevant supported entry types</a>. The user
+            agent SHOULD notify developers if <var>entry types</var> is
+            modified. For example, a console warning listing removed types might
+            be appropriate.</li>
             <li>If the resulting <var>entry types</var> sequence is an empty
             sequence, abort these steps. The user agent SHOULD notify developers
             when the steps are aborted to notify that registration has been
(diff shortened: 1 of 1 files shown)


npm1:
I'd like to find an easier solution that hopefully does not involve having both a list and a FrozenArray :)

npm1:
@bzbarsky tagging you as you cannot be added as a reviewer. What do you think? In tandem other specifications will need to be modified to include something such as:

> The <a>frozen supported entry types</a> of all window environments MUST contain the string 'paint'.

npm1:
Tagging relevant issues: https://github.com/w3c/performance-timeline/issues/113 https://github.com/w3c/performance-timeline/issues/117

index.html
```diff
@@ -385,10 +397,10 @@ <h2><dfn>observe()</dfn> method</h2>
             <a>entryTypes</a> sequence.
             </li>
             <li>Remove all types from <var>entry types</var> that are not
-            contained in <a>supported entry types</a>. The user agent SHOULD
-            notify developers if <var>entry types</var> is modified. For
-            example, a console warning listing removed types might be
-            appropriate.</li>
+            contained in the <a>relevant supported entry types</a>. The user
```
bzbarsky: Two issues here:

1) This is not well-defined, unfortunately.  "relevant supported entry types" is defined in terms of "relevant global object", but without defining _what_ it's a relevant global object of.  "relevant global object" is defined for platform objects, not as an abstract concept.  Concretely, the definition of "relevant supported entry types" should include an input: the platform object it's for.  And then it can pass that input through to the "relevant global object" concept.  So in this case you'd say "the relevant supported entry types of _observer_" and the definition of the "relevant supported entry types" would talk about the "relevant supported entry types for a platform object _o_ are ... of the relevant global object of _o_".

2) The definition of "contained in" is not clear here, given that the "relevant supported entry types" is a JS Array.  How is that determination made?  Is the method observable via side-effects on the page?  Ideally we would sidestep this by storing both an infra list and the JS Array representation of it on the global and just consult the list here.

index.html
```diff
@@ -559,19 +571,10 @@ <h2><dfn>disconnect()</dfn> method</h2>
...
-        specifications that define new entry types via the
-        <a>register a performance entry type</a> algorithm.</p>
-        <p>When <dfn>supportedEntryTypes</dfn>'s attribute getter is called, run the following steps:</p>
-        <ol>
-          <li>Let <var>result</var> be a copy of the user agent's <a>supported entry types</a>.</li>
-          <li>Sort <var>result</var> in alphabetical order.</li>
-          <li>Return <var>result</var>.</li>
-        </ol>
+        <p>When <dfn>supportedEntryTypes</dfn>'s attribute getter is called, it
+          MUST return the <a>relevant supported entry types</a>.</p>
```
bzbarsky: This doesn't work, unfortunately, because "relevant supported entry types" as defined above is only defined with reference to a platform object, but there is no platform object present here at all: it's a static attribute getter.

index.html
```diff
@@ -193,17 +193,29 @@ <h2><dfn>Performance Timeline</dfn></h2>
...
-      <li>a <dfn>performance entry buffer</dfn> to store
-      <a>PerformanceEntry</a> objects that is initially empty
+      <li>A <dfn>performance observer task queued flag</dfn>.</li>
+      <li>A <dfn>list of <a>registered performance observer</a> objects</dfn>
+      that is initially empty.</li>
+      <li>A <dfn>performance entry buffer</dfn> to store
+      <a>PerformanceEntry</a> objects that is initially empty.
+      </li>
+      <li>
+        A <dfn>frozen supported entry types</dfn> representing the cached
```
rniwa: Let's remove the word "cached" here since that's not really a concept in WebIDL.

index.html

index.html
npm1: Improved the definition and added the infra list.

npm1:
This is ready for review again, sorry for the delay.

By the way, @domenic suggested having a complete registration list in PerformanceTimeline instead of having a registry that other specfications hook into (which is the current approach). The registration list approach seems to be easier to spec and more in line with what an implementation would actually do, but @yoavweiss opposed because there would not be a clear way to signal implementors of new specifications that they should update the list, the registration seems incompatible with W3C workmode, and testing would be trickier (we want tests for specific strings to be in the folder of the corresponding specification, for example 'longtask' in the Long Tasks folder). Surfacing this in case people here have opinions!

rniwa:
I think having a hook and each specification adding its own entry makes more sense to me. A centralized list would be rather misleading if some browsers didn't support certain specs, etc...

npm1:
> I think having a hook and each specification adding its own entry makes more sense to me. A centralized list would be rather misleading if some browsers didn't support certain specs, etc...

Hmm well, this was discussed on the latest call and the decision was to have a centralized list but outside of PerformanceTimeline spec. You can state your disagreement in https://github.com/w3c/performance-timeline/issues/117, though.

npm1:
Closing this since we're not following this approach. Will do another PR once https://github.com/w3c/performance-timeline/pull/132 lands

Issue #119, opened 2019-03-28
WebIDL spec does not allow required dictionary argument with no required members


hawkinsw:
It is my understanding from reading the WebIDL spec that the combination of the proposed declaration of the observe() method and the PerformanceObserverInit dictionary are incompatible. 

In particular, the [spec](https://heycam.github.io/webidl/#idl-operations) says that 

> If the type of an argument is a dictionary type or a union type that has a dictionary as one of its flattened member types, and that dictionary type and its ancestors have no required members, and the argument is either the final argument or is followed only by optional arguments, then the argument must be specified as optional. Such arguments are always considered to have a default value of an empty dictionary, unless otherwise specified.

There are precedents with other specs where this same change needed to be made. E.g., https://github.com/whatwg/dom/issues/316

A change to `optional` will have a knock-on affect as to the value of the `observe` method's `length`. (It will go from 1 to 0). 

I hope that this is a cogent analysis and look forward to helping make any necessary changes. Thank you!

npm1:
Ahh thanks for the investigation! So I think the only change needed here would be to make the `PerformanceObserverInit options` optional. The observe processing model already throws a SyntaxError if the `entryTypes` and `type` members are both missing, so it shouldn't affect behavior.

@igrigorik @yoavweiss @toddreifsteck @rniwa @mstange FYI

whh8b:
Thanks for the reply. I agree with you.

rniwa:
Makes sense.

PR #120, opened 2019-03-28
Make observe() parameter an optional dictionary

npm1:
It seems that it must be made optional to confirm with WebIDL spec. Fortunately the observe() processing will already throw if 'entryTypes' and 'type' are both omitted (which is what would happen if the parameter is missing, as an empty dictionary is assumed), so no additional changes are required.
Solves https://github.com/w3c/performance-timeline/issues/119
@whh8b FYI

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/pull/120.html" title="Last updated on Mar 28, 2019, 4:21 PM UTC (c93fbf8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/120/3139cce...c93fbf8.html" title="Last updated on Mar 28, 2019, 4:21 PM UTC (c93fbf8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -323,7 +323,7 @@
                                                    PerformanceObserver observer);
       [Constructor(PerformanceObserverCallback callback), Exposed=(Window,Worker)]
       interface PerformanceObserver {
-        void observe (PerformanceObserverInit options);
+        void observe (optional PerformanceObserverInit options);
         void disconnect ();
         PerformanceEntryList takeRecords();
         static readonly attribute FrozenArray&lt;DOMString&gt; supportedEntryTypes;


whh8b:
Note: I think that we will have to modify the test suites as well. The tests expect that PerformanceObserver.observe.length returns 1 (for the required parameter) but will now return 0 in this case. At least I *think*.

npm1:
Do you mean the IDL harness test? That one is automatically generated from the spec.

whh8b:
That's the one! I didn't realize that it was automatic. Thanks for teaching
me that. Sorry for the useless note.

PR #123, opened 2019-04-03
Buffered flag spec changes

ehanley324:
https://github.com/w3c/performance-timeline/issues/81

***
<a href="https://pr-preview.s3.amazonaws.com/ehanley324/performance-timeline/pull/123.html" title="Last updated on May 9, 2019, 3:06 PM UTC (04441b7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/123/58b1f72...ehanley324:04441b7.html" title="Last updated on May 9, 2019, 3:06 PM UTC (04441b7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -196,14 +196,14 @@
       <li>a <dfn>performance observer task queued flag</dfn></li>
       <li>a <dfn>list of <a>registered performance observer</a> objects</dfn>
       that is initially empty</li>
-      <li>a <dfn>performance entry buffer</dfn> per observed
-      <dfn>entryType</dfn> to store <a>PerformanceEntry</a> objects that
+      <li>a <dfn>performance entry buffer map</dfn>, one per
+      <dfn>entryType</dfn>, to store <a>PerformanceEntry</a> objects, that
       are initially empty.  Each entryType's buffer has a
       <dfn>maxBufferSize</dfn> that with rare exception, cannot be changed.
       </li>
     </ul>
-    <p>The <dfn>relevant performance entry buffer</dfn> is the <a>performance
-    entry buffer</a> associated with the <a data-cite=
+    <p>The <dfn>relevant performance entry buffer</dfn> is the object in the
+    <a>performance entry buffer map</a> associated with the <a data-cite=
     "HTML/webappapis.html#concept-relevant-global">relevant global
     object</a> for a specific <dfn>entryType</dfn>.</p>
     <section data-dfn-for="Performance" data-link-for="Performance">
@@ -242,7 +242,7 @@
       <section>
         <h2><dfn>getEntriesByName()</dfn> method</h2>
         <p>Returns a <a>PerformanceEntryList</a> object returned by <a href=
-        "#filter-buffers-by-name-and-type"></a> algorithm with <var>buffer</var>
+        "#filter-buffers-by-name-and-type"></a> algorithm with <var>buffers</var>
         set to <a>relevant performance entry buffer</a> for every type,
         <var>name</var> set to `name`, and <var>type</var> set to
         <code>null</code> if optional
@@ -508,7 +508,7 @@
           <dl>
             <dt><dfn>buffered</dfn></dt>
             <dd>A flag to indicate whether buffered entries should be queued
-            into observer's buffer for each entry.</dd>
+            into observer's buffer.</dd>
           </dl>
         </section>
         <section data-dfn-for="PerformanceObserverEntryList" data-link-for=
@@ -610,8 +610,7 @@
         <li>For each <var>observer</var> in <var>interested observers</var>:
           <ol>
             <li>Append <var>new entry</var> to <var>observer</var>'s
-            <a>observer buffer</a> for that <a data-lt=
-              "PerformanceEntry.entryType">entryType</a>.
+            <a>observer buffer</a>.
             </li>
           </ol>
         </li>
--- a/index.html
+++ b/index.html
@@ -196,14 +196,16 @@
       <li>a <dfn>performance observer task queued flag</dfn></li>
       <li>a <dfn>list of <a>registered performance observer</a> objects</dfn>
       that is initially empty</li>
-      <li>a <dfn>performance entry buffer</dfn> to store
-      <a>PerformanceEntry</a> objects that is initially empty
+      <li>a <dfn>performance entry buffer</dfn> per observed
+      <dfn>entryType</dfn> to store <a>PerformanceEntry</a> objects that
+      are initially empty.  Each entryType's buffer has a
+      <dfn>maxBufferSize</dfn> that with rare exception, cannot be changed.
       </li>
     </ul>
     <p>The <dfn>relevant performance entry buffer</dfn> is the <a>performance
     entry buffer</a> associated with the <a data-cite=
     "HTML/webappapis.html#concept-relevant-global">relevant global
-    object</a>.</p>
+    object</a> for a specific <dfn>entryType</dfn>.</p>
     <section data-dfn-for="Performance" data-link-for="Performance">
       <h2>Extensions to the <dfn data-cite=
       "hr-time-2#sec-performance">Performance</dfn> interface</h2>
@@ -225,23 +227,25 @@
       <section>
         <h2><dfn>getEntries()</dfn> method</h2>
         <p>Returns a <a>PerformanceEntryList</a> object returned by <a href=
-        "#filter-buffer-by-name-and-type"></a> algorithm with <var>buffer</var>
-        set to <a>relevant performance entry buffer</a>, and <var>name</var>
-        and <var>type</var> set to <code>null</code>.</p>
+        "#filter-buffers-by-name-and-type"></a> algorithm with <var>buffers</var>
+        set to <a>relevant performance entry buffer</a> for every type,
+        and <var>name</var> and <var>type</var> set to <code>null</code>.</p>
       </section>
       <section>
         <h2><dfn>getEntriesByType()</dfn> method</h2>
         <p>Returns a <a>PerformanceEntryList</a> object returned by <a href=
-        "#filter-buffer-by-name-and-type"></a> algorithm with <var>buffer</var>
-        set to <a>relevant performance entry buffer</a>, <var>name</var> set to
-        <code>null</code>, and <var>type</var> set to `type`.</p>
+        "#filter-buffers-by-name-and-type"></a> algorithm with <var>buffer</var>
+        set to <a>relevant performance entry buffer</a> for `type`,
+        <var>name</var> set to <code>null</code>, and <var>type</var>
+        set to `type`.</p>
       </section>
       <section>
         <h2><dfn>getEntriesByName()</dfn> method</h2>
         <p>Returns a <a>PerformanceEntryList</a> object returned by <a href=
-        "#filter-buffer-by-name-and-type"></a> algorithm with <var>buffer</var>
-        set to <a>relevant performance entry buffer</a>, <var>name</var> set to
-        `name`, and <var>type</var> set to <code>null</code> if optional
+        "#filter-buffers-by-name-and-type"></a> algorithm with <var>buffer</var>
+        set to <a>relevant performance entry buffer</a> for every type,
+        <var>name</var> set to `name`, and <var>type</var> set to
+        <code>null</code> if optional
         `entryType` is omitted, and <var>type</var> set to `type`
         otherwise.</p>
       </section>
@@ -275,7 +279,9 @@
           <code>"measure"</code> [[USER-TIMING-2]], <code>"navigation"</code>
           [[NAVIGATION-TIMING-2]], <code>"resource"</code>
           [[RESOURCE-TIMING-2]], <!-- TODO: add long task spec reference -->
-           and <code>"longtask"</code>.</p>
+           and <code>"longtask"</code>.  These specifications need to indicate
+          whether the specifiy entryType is available from the
+          <a>Performance</a> interface.</p>
         </dd>
         <dt><dfn>startTime</dfn></dt>
         <dd>This attribute MUST return the time value of the first recorded
@@ -444,7 +450,7 @@
             <li>If <var>options</var>'s <a>buffered</a> flag is set:
               <ol>
                 <li>Let <var>entries</var> be the <a>PerformanceEntryList</a>
-                object returned by <a href="#filter-buffer-by-name-and-type"></a>
+                object returned by <a href="#filter-buffers-by-name-and-type"></a>
                 algorithm with <var>buffer</var> set to <a>relevant
                 performance entry buffer</a>, <var>name</var> set to
                 <code>null</code> and <var>type</var> set to
@@ -502,7 +508,7 @@
           <dl>
             <dt><dfn>buffered</dfn></dt>
             <dd>A flag to indicate whether buffered entries should be queued
-            into observer's buffer.</dd>
+            into observer's buffer for each entry.</dd>
           </dl>
         </section>
         <section data-dfn-for="PerformanceObserverEntryList" data-link-for=
@@ -519,21 +525,21 @@
           <section>
             <h2><dfn>getEntries()</dfn> method</h2>
             <p>Returns a <a>PerformanceEntryList</a> object returned by
-            <a href="#filter-buffer-by-name-and-type"></a> algorithm with
+            <a href="#filter-buffers-by-name-and-type"></a> algorithm with
             <var>buffer</var> set to <a>observer buffer</a>, and
             <var>name</var> and <var>type</var> set to <code>null</code>.</p>
           </section>
           <section>
             <h2><dfn>getEntriesByType()</dfn> method</h2>
             <p>Returns a <a>PerformanceEntryList</a> object returned by
-            <a href="#filter-buffer-by-name-and-type"></a> algorithm with
+            <a href="#filter-buffers-by-name-and-type"></a> algorithm with
             <var>buffer</var> set to <a>observer buffer</a>, <var>name</var>
             set to <code>null</code>, and <var>type</var> set to `type`.</p>
           </section>
           <section>
             <h2><dfn>getEntriesByName()</dfn> method</h2>
             <p>Returns a <a>PerformanceEntryList</a> object returned by
-            <a href="#filter-buffer-by-name-and-type"></a> algorithm with
+            <a href="#filter-buffers-by-name-and-type"></a> algorithm with
             <var>buffer</var> set to <a>observer buffer</a>, <var>name</var>
             set to `name`, and <var>type</var> set to <code>null</code> if
             optional `entryType` is omitted, and <var>type</var> set to `type`
@@ -581,9 +587,8 @@
     <h2>Processing</h2>
     <section data-link-for="PerformanceObserver">
       <h2>Queue a <code>PerformanceEntry</code></h2>
-      <p>To <dfn>queue a PerformanceEntry</dfn> (<var>new entry</var>) with an
-      optional <var>add to performance entry buffer</var> flag, which is unset
-      by default, run these steps:</p>
+      <p>To <dfn>queue a PerformanceEntry</dfn> (<var>new entry</var>), 
+      run these steps:</p>
       <ol>
         <li>Let <var>interested observers</var> be an initially empty set of
         <a>PerformanceObserver</a> objects.
@@ -605,12 +610,14 @@
         <li>For each <var>observer</var> in <var>interested observers</var>:
           <ol>
             <li>Append <var>new entry</var> to <var>observer</var>'s
-            <a>observer buffer</a>.
+            <a>observer buffer</a> for that <a data-lt=
+              "PerformanceEntry.entryType">entryType</a>.
             </li>
           </ol>
         </li>
-        <li>If the <var>add to performance entry buffer</var> flag is set, add
-        <var>new entry</var> to the <a>performance entry buffer</a>.
+        <li>If the <a>maxBufferSize</a> for this <var>entryType</var> has not
+        been reached, add <var>new entry</var> to the
+        <a>performance entry buffer</a> for that <var>entryType</var>.
         </li>
         <li>If the <a>performance observer task queued flag</a> is set,
         terminate these steps.
@@ -660,8 +667,8 @@
       periods to minimize impact of performance monitoring code.</p>
     </section>
     <section data-link-for="PerformanceEntry">
-      <h2>Filter <var>buffer</var> by <var>name</var> and <var>type</var></h2>
-      <p>Given a <var>buffer</var> and optional <var>name</var> and
+      <h2>Filter <var>buffers</var> by <var>name</var> and <var>type</var></h2>
+      <p>Given a list of <var>buffers</var> and optional <var>name</var> and
       <var>type</var> string values this algorithm returns a
       <a>PerformanceEntryList</a> object that contains a list of
       <a>PerformanceEntry</a> objects sorted in chronological order with
@@ -672,7 +679,7 @@
         <a>PerformanceEntryList</a>.
         </li>
         <li>For each <a>PerformanceEntry</a> object (<var>entryObject</var>) in
-        the <var>buffer</var>, in chronological order with respect to
+        the <var>buffers</var>, in chronological order with respect to
         <a>startTime</a>:
           <ol>
             <li>If <var>name</var> is not <code>null</code> and
--- a/index.html
+++ b/index.html
@@ -205,7 +205,7 @@
           <li> A <code>long</code> indicating the <dfn>maxBufferSize</dfn> for
             this entryType.  With rare exception, this value cannot be changed
             once set</li>
-          <li> A <code>boolean</code> <dfn>availableFromTimeline<dfn>
+          <li> A <code>boolean</code> <dfn>availableFromTimeline</dfn>
             representing whether this entryType is available from the
             <a>Performance</a> interface.</li>
         </ul>
@@ -241,14 +241,14 @@
         "#filter-buffers-by-name-and-type"></a> algorithm with
           <ol>
             <li>
-              <var>buffers</var>set to a list containing
+              <var>buffers</var> set to a list containing
               <a>relevant performance entry buffer</a>s where
               <var>entryType</var> is set to each entry type supported by the
               user agent and whose <a>availableFromTimeline</a> boolean is set
               to true
             </li>
-              <var>name</var> set to <code>null</code>.
             <li>
+              <var>name</var> set to <code>null</code>.
             </li>
             <li>
               <var>type</var> set to <code>null</code>.
@@ -262,13 +262,13 @@
         "#filter-buffers-by-name-and-type"></a> algorithm with:
           <ol>
             <li>
-              <var>buffers</var>set to a list containing
+              <var>buffers</var> set to a list containing
               <a>relevant performance entry buffer</a>s where
               <var>entryType</var> is set to `type` if the associated
-              <a>availableFromeTimeline</a> boolean is set to true
+              <a>availableFromTimeline</a> boolean is set to true
             </li>
-              <var>name</var> set to <code>null</code>.
             <li>
+              <var>name</var> set to <code>null</code>.
             </li>
             <li>
               <var>type</var> set to `type`.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -196,14 +196,16 @@ <h2><dfn>Performance Timeline</dfn></h2>
       <li>a <dfn>performance observer task queued flag</dfn></li>
       <li>a <dfn>list of <a>registered performance observer</a> objects</dfn>
       that is initially empty</li>
-      <li>a <dfn>performance entry buffer</dfn> to store
-      <a>PerformanceEntry</a> objects that is initially empty
+      <li>a <dfn>performance entry buffer</dfn> per observed
```
npm1: What does 'observerd entryType' mean? Maybe the correct way to do this would be to say that other specs that define new entry types need to specify a new buffer for that type. And here we have a map entryType->buffer.

index.html
```diff
@@ -225,23 +227,25 @@ <h2>Extensions to the <dfn data-cite=
...
+        set to <a>relevant performance entry buffer</a> for every type,
+        and <var>name</var> and <var>type</var> set to <code>null</code>.</p>
       </section>
       <section>
         <h2><dfn>getEntriesByType()</dfn> method</h2>
         <p>Returns a <a>PerformanceEntryList</a> object returned by <a href=
-        "#filter-buffer-by-name-and-type"></a> algorithm with <var>buffer</var>
-        set to <a>relevant performance entry buffer</a>, <var>name</var> set to
-        <code>null</code>, and <var>type</var> set to `type`.</p>
+        "#filter-buffers-by-name-and-type"></a> algorithm with <var>buffer</var>
```
npm1: Should it be 'buffers' here too, and below?

index.html
```diff
@@ -275,7 +279,9 @@ <h2>The <dfn>PerformanceEntry</dfn> interface</h2>
           <code>"measure"</code> [[USER-TIMING-2]], <code>"navigation"</code>
           [[NAVIGATION-TIMING-2]], <code>"resource"</code>
           [[RESOURCE-TIMING-2]], <!-- TODO: add long task spec reference -->
-           and <code>"longtask"</code>.</p>
+           and <code>"longtask"</code>.  These specifications need to indicate
```
npm1: This seems like the wrong place for this comment.

index.html
```diff
@@ -444,7 +450,7 @@ <h2><dfn>observe()</dfn> method</h2>
             <li>If <var>options</var>'s <a>buffered</a> flag is set:
               <ol>
                 <li>Let <var>entries</var> be the <a>PerformanceEntryList</a>
-                object returned by <a href="#filter-buffer-by-name-and-type"></a>
+                object returned by <a href="#filter-buffers-by-name-and-type"></a>
                 algorithm with <var>buffer</var> set to <a>relevant
```
npm1: Again here we would like consistency with the change in the input to this algorithm.

index.html
```diff
@@ -502,7 +508,7 @@ <h2><dfn>PerformanceObserverInit</dfn> dictionary</h2>
           <dl>
             <dt><dfn>buffered</dfn></dt>
             <dd>A flag to indicate whether buffered entries should be queued
-            into observer's buffer.</dd>
+            into observer's buffer for each entry.</dd>
```
npm1: Not sure what this addition is trying to clarify.

index.html
```diff
@@ -605,12 +610,14 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
         <li>For each <var>observer</var> in <var>interested observers</var>:
           <ol>
             <li>Append <var>new entry</var> to <var>observer</var>'s
-            <a>observer buffer</a>.
+            <a>observer buffer</a> for that <a data-lt=
```
npm1: This seems wrong. A single PerformanceObserver still has only one buffer, which contains all of the entries, no?

index.html
ehanley324: Yeah observed didn't belong here.  

I updated to map.  Not clear what you mean by "other specs that define new entry types need to specify a new buffer for that type".  This map will contain buffers for all entryTypes, we don't obtain those buffers elsewhere.

index.html
ehanley324: getEntriesByType is only ever one type, so one buffer.  getEntriesByName should indicate buffers since we are looking in potentially all buffers.

index.html
ehanley324: Again this is only a single type, so I think it should be singular.

index.html
ehanley324: Right, done.

index.html
ehanley324: Like below, at some point I was thinking of the observers buffer differently. Removed.

index.html
ehanley324: I dont disagree.  This is talking about the attribute in the PerformanceEntry object, but this is the only place in the spec where we call out where these types come from and it is these specs that need to determine if they are available from the timeline. 

Do you have another suggestion?  I think maybe it should be in the "Extensions to the Prformance Interface" section?  It isn't a method specifically, but maybe just add a "Note" in the 3.1 section?

index.html
ehanley324: Nicolas and I chatted offline about making sure the variables line up this should be resolved.

index.html
ehanley324: We spoke offline and this is now updated.

index.html
ehanley324: I updated this map again after we spoke offline.  PTAL

index.html
```diff
@@ -196,16 +196,25 @@ <h2><dfn>Performance Timeline</dfn></h2>
...
+            representing whether this entryType is available from the
+            <a>Performance</a> interface.</li>
+        </ul>
       </li>
     </ul>
-    <p>The <dfn>relevant performance entry buffer</dfn> is the object in the
-    <a>performance entry buffer map</a> associated with the <a data-cite=
-    "HTML/webappapis.html#concept-relevant-global">relevant global
-    object</a> for a specific <dfn>entryType</dfn>.</p>
+    <p>The <dfn>relevant performance entry buffer</dfn> is the <a>performance
```
npm1: This dfn now needs an entryType as a parameter in order to make sense.

index.html
```diff
@@ -228,22 +237,28 @@ <h2>Extensions to the <dfn data-cite=
...
+        each entry type supported by the user agent and whose
+        <a>availableFromeTimeline</a> boolean is set to true and <var>name</var>
+        and <var>type</var> set to <code>null</code>.</p>
       </section>
       <section>
         <h2><dfn>getEntriesByType()</dfn> method</h2>
         <p>Returns a <a>PerformanceEntryList</a> object returned by <a href=
-        "#filter-buffers-by-name-and-type"></a> algorithm with <var>buffer</var>
-        set to <a>relevant performance entry buffer</a> for `type`,
+        "#filter-buffers-by-name-and-type"></a> algorithm with
```
npm1: This is too complex now so should be converted to an algorithm.

index.html
```diff
@@ -228,22 +237,28 @@ <h2>Extensions to the <dfn data-cite=
...
+        <a>availableFromeTimeline</a> boolean is set to true,
         <var>name</var> set to <code>null</code>, and <var>type</var>
         set to `type`.</p>
       </section>
       <section>
         <h2><dfn>getEntriesByName()</dfn> method</h2>
         <p>Returns a <a>PerformanceEntryList</a> object returned by <a href=
         "#filter-buffers-by-name-and-type"></a> algorithm with <var>buffers</var>
-        set to <a>relevant performance entry buffer</a> for every type,
+        set to a list containing a <a>relevant performance entry buffer</a> for
```
npm1: ditto

index.html
```diff
@@ -451,8 +464,8 @@ <h2><dfn>observe()</dfn> method</h2>
               <ol>
                 <li>Let <var>entries</var> be the <a>PerformanceEntryList</a>
                 object returned by <a href="#filter-buffers-by-name-and-type"></a>
-                algorithm with <var>buffer</var> set to <a>relevant
-                performance entry buffer</a>, <var>name</var> set to
+                algorithm with <var>buffers</var> set to a list containing the
+                <a>relevant performance entry buffer</a>, <var>name</var> set to
```
npm1: Need to specify the entryType.

index.html
```diff
@@ -228,22 +237,28 @@ <h2>Extensions to the <dfn data-cite=
...
       </section>
       <section>
         <h2><dfn>getEntriesByType()</dfn> method</h2>
         <p>Returns a <a>PerformanceEntryList</a> object returned by <a href=
-        "#filter-buffers-by-name-and-type"></a> algorithm with <var>buffer</var>
-        set to <a>relevant performance entry buffer</a> for `type`,
+        "#filter-buffers-by-name-and-type"></a> algorithm with
+        <var>buffers</var> set to a list containing a <a>
+        relevant performance entry buffer</a> for `type` if the associated
+        <a>availableFromeTimeline</a> boolean is set to true,
```
npm1: Typo 'availableFromeTimeline'

index.html
npm1: For clarity: ditto, should be converted to algorithm

index.html
```diff
@@ -236,33 +238,65 @@ <h2>Extensions to the <dfn data-cite=
       <section>
         <h2><dfn>getEntries()</dfn> method</h2>
         <p>Returns a <a>PerformanceEntryList</a> object returned by <a href=
-        "#filter-buffers-by-name-and-type"></a> algorithm with <var>buffers</var>
-        set to a list containing a <a>relevant performance entry buffer</a> for
-        each entry type supported by the user agent and whose
-        <a>availableFromeTimeline</a> boolean is set to true and <var>name</var>
-        and <var>type</var> set to <code>null</code>.</p>
+        "#filter-buffers-by-name-and-type"></a> algorithm with
```
npm1: Missing ':'
with:

index.html
```diff
@@ -236,33 +238,65 @@ <h2>Extensions to the <dfn data-cite=
...
+          <ol>
+            <li>
+              <var>buffers</var> set to a list containing
```
npm1: I think it is better to split this out into a new algorithm which is similar to filter-buffers-by-name-and-type, except that it excludes entries for which availableFromTimeline is false. This would having this structure 3 times. What do you think?

index.html
npm1: Maybe we can simplify it somehow to make it less confusing. But there is not a single performance entry buffer now. So that did not work as it was.

npm1:
Adding @yoavweiss for an additional set of eyes and maybe ideas to help simplifying some of the logic here.

index.html
yoavweiss: Agree with @npm1 here. We'd want each spec to explicitly "register" its buffer, in a similar way to the way each spec is [registering supported entry types](https://w3c.github.io/performance-timeline//#dfn-register-a-performance-entry-type). We need that registration to include the full tuple, so indicate a) its entryType b) its max size c) its visibility from the timeline

index.html
```diff
@@ -609,8 +664,9 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
             </li>
           </ol>
         </li>
-        <li>If the <var>add to performance entry buffer</var> flag is set, add
-        <var>new entry</var> to the <a>performance entry buffer</a>.
+        <li>If the <a>maxBufferSize</a> for this <var>entryType</var> has not
```
yoavweiss: This is underdefined. You need to use the entryType to fetch the right tuple from the map, get maxBufferSize from that and then use that.

ehanley324: The new version includes these updates to refer to the map.

index.html
ehanley324: Done with Yoav's changes.

index.html
ehanley324: resolved with Yoav's changes

index.html
ehanley324: Is this something I need to explicitly call out somewhere that each spec needs to do this?

yoavweiss: You need to add an algorithm that performs that registration here for a generic entry type, and each spec would then need to call into that algorithm to register its buffer.

yoavweiss:
General comment to make the IPR bots happy: if you don't have one yet, you'd need to create a W3C account, link it with your GH account and join the WebPerfWG.

index.html
```diff
@@ -609,8 +639,24 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
...
         </li>
-        <li>If the <var>add to performance entry buffer</var> flag is set, add
-        <var>new entry</var> to the <a>performance entry buffer</a>.
+        <li>
+          Determine if you should add <var>new entry</var> to a 
+          <a>performance entry buffer</a>:
+          <ol>
+            <li>Let <var>map</var> be the <a>performance entry buffer map</a>
+              associated with the <a
+              data-cite="HTML/webappapis.html#concept-relevant-global"> relevant
```
yoavweiss: Nit: extra space

index.html
```diff
@@ -609,8 +639,24 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
             </li>
           </ol>
         </li>
-        <li>If the <var>add to performance entry buffer</var> flag is set, add
-        <var>new entry</var> to the <a>performance entry buffer</a>.
+        <li>
+          Determine if you should add <var>new entry</var> to a
```
yoavweiss: This seems like something you'd add as a code comment. In a spec, there are a couple of ways to do that: a) Define it as an independent algorithm which name states its use (kinda like helper functions with a meaningful name) b) Add it as a note

index.html
```diff
@@ -609,8 +639,24 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
             </li>
           </ol>
         </li>
-        <li>If the <var>add to performance entry buffer</var> flag is set, add
-        <var>new entry</var> to the <a>performance entry buffer</a>.
+        <li>
+          Determine if you should add <var>new entry</var> to a 
+          <a>performance entry buffer</a>:
+          <ol>
+            <li>Let <var>map</var> be the <a>performance entry buffer map</a>
```
yoavweiss: This seems like a good place to call the "get the relevant performance entry tuple" algorithm

index.html
```diff
@@ -443,18 +466,22 @@ <h2><dfn>observe()</dfn> method</h2>
...
-                performance entry buffer</a>, <var>name</var> set to
-                <code>null</code> and <var>type</var> set to
-                <var>options</var>'s <a>type</a>.
-                </li>
-                <li>Append <var>entries</var> to the <a data-cite=
-                "WHATWG-DOM#context-object">context object</a>'s <a>observer
-                buffer</a>.
-                </li>
-              </ol>
+                <li>Let <var>map</var> be the <a>performance entry buffer map</a>
```
yoavweiss: Here as well, you can call the "get the relevant tuple" algorithm defined above instead if accessing the map directly

index.html
ehanley324: So it is more like the previous statement when I used the "get the releveant performance entry tuple" algorithm.  Let me know if you think it still makes more sense as a comment.  To me it is a pretty central part of queuing a PerformanceEntry object.

index.html
ehanley324: Done.

index.html
ehanley324: Done.

index.html
```diff
@@ -609,8 +634,11 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
             </li>
           </ol>
         </li>
-        <li>If the <var>add to performance entry buffer</var> flag is set, add
-        <var>new entry</var> to the <a>performance entry buffer</a>.
+        <li>
+          Let <var>tuple</var> be the <a>relevant performance entry tuple</a>,
```
yoavweiss: I think it'll be clearer to split that up into 2 steps. Also, the max buffer limitation should be clearer, to avoid off-by-one errors.

Maybe something like:
* Let tuple be the relevant.. tuple
* If the number of entries in the tuple's performance entry buffer is smaller than tuple's maxBufferSize, append...

In ResourceTiming, we explicitly created variables that count the number of entries, but I think a "number of entries" language would be precise enough

index.html
```diff
@@ -660,39 +688,69 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
...
-        <a>PerformanceEntryList</a>.
-        </li>
-        <li>For each <a>PerformanceEntry</a> object (<var>entryObject</var>) in
-        the <var>buffer</var>, in chronological order with respect to
-        <a>startTime</a>:
+        <li>Let <var>result</var> be an initially empty <a
+        data-cite="WHATWG-INFRA#list">list</a>.</li>
+        <li>Let <var>map</var> be the <a>performance entry buffer map</a>
+        associated with the <a
+        data-cite="HTML/webappapis.html#concept-relevant-global"> relevant
```
yoavweiss: Nit: spurious space

index.html
```diff
@@ -660,39 +688,69 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
...
+            buffer by name and type</a> with <var>buffer</var>, <var>name</var>
+            and <var>type</var> as inputs.</li>
+            <li>For each <a>PerformanceEntry</a> <var>entry</var> in
+            <var>entries</var>, <a
+            data-cite="WHATWG-INFRA#list-append">append</a> it to
+            <var>result</var>.</li>
           </ol>
         </li>
-        <li>Return the <var>list of entry objects</var>.
+        <li>Return <var>result</var>.</li>
```
yoavweiss: My bad, but I think we'd need to sort `result` here before returning it, if we want this to be in line with the previous spec language. We'd also need to add test to prove that this is indeed what's currently implemented.

index.html
ehanley324: Done.  And again I think this should remain here for two reasons: 
1) It is more central information than a note, it is part of the queuing algorithm for a PerformanceEntry
2) It is not re-used anywhere else in the spec to be a common algorithm

index.html
ehanley324: Done.  Yes this is how it is implemented in chrome.

index.html
```diff
@@ -443,18 +465,18 @@ <h2><dfn>observe()</dfn> method</h2>
             as its only item.</li>
             <li>If <var>options</var>'s <a>buffered</a> flag is set:
               <ol>
-                <li>Let <var>entries</var> be the <a>PerformanceEntryList</a>
-                object returned by <a href="#filter-buffer-by-name-and-type"></a>
-                algorithm with <var>buffer</var> set to <a>relevant
-                performance entry buffer</a>, <var>name</var> set to
-                <code>null</code> and <var>type</var> set to
-                <var>options</var>'s <a>type</a>.
+                <li>
```
yoavweiss: Nit: spurious new line

index.html
```diff
@@ -443,18 +465,18 @@ <h2><dfn>observe()</dfn> method</h2>
...
-                <li>Append <var>entries</var> to the <a data-cite=
-                "WHATWG-DOM#context-object">context object</a>'s <a>observer
-                buffer</a>.
-                </li>
-              </ol>
+                <li>For each <a>PerformanceEntry</a> <var>entry</var> in
+                  <var>tuple</var>'s <a>performance entry buffer</a>
+                  <a data-cite="WHATWG-INFRA#list-append">append</a> it to
+                  the <a data-cite=
+                  "WHATWG-DOM#context-object">context object</a>'s <a>observer
```
yoavweiss: Re-reviewing this, it would be clearer to say `observer` here, rather than refer (again) to the context object.

index.html
```diff
@@ -609,8 +634,18 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
             </li>
           </ol>
         </li>
-        <li>If the <var>add to performance entry buffer</var> flag is set, add
-        <var>new entry</var> to the <a>performance entry buffer</a>.
+        <li>
+          Determine if you should add <var>new entry</var> to a
```
yoavweiss: You should think of this algorithm as you think of code. You wouldn't tell a computer to "Determine if you should add", as that is not defined. What you can do here is:
a) Add that as a note, which is an equivalent to a code comment
b) Split out this sub-algorithm to a "helper function". You can add a separate algorithm below that defines what it means to "determine if you should add" and call it from here.

Does that make sense?

index.html
```diff
@@ -609,8 +634,18 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
...
+          <a>performance entry buffer</a>:
+          <ol>
+            <li> Let <var>tuple</var> be the <a>relevant performance entry
```
yoavweiss: Nit: Spurious space, missing period at end of line.

index.html
```diff
@@ -660,39 +695,71 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
...
+        respect to <a>startTime</a></li>
+        <li>Return <var>result</var>.</li>
+      </ol>
+      <p>When asked to run the <dfn>filter buffer by name and type</dfn>
+      algorithm, with <var>buffer</var>, <var>name</var>, and <var>type</var>
+      as inputs, run the following steps:</p>
+      <ol>
+        <li>Let <var>result</var> be an initially empty <a
+        data-cite="WHATWG-INFRA#list">list</a>.</li>
+        <li>Sort <var>buffer</var>'s entries in chronological order with
```
yoavweiss: I think it's better to sort the results list here, rather than the underlying buffer.

yoavweiss:
@ehanley324 - in order to appease the IPR bots, could you join the WG and connect your GH account with your W3C one? Thanks! :)

index.html
ehanley324: Done.

index.html
ehanley324: Done.

ehanley324:
> @ehanley324 - in order to appease the IPR bots, could you join the WG and connect your GH account with your W3C one? Thanks! :)

Done.

index.html
yoavweiss: To better explain my comment, which I now see is not very clear. It would be better to move the sorting step until after the loop is run on each entry in the buffer, and then sort `results`

index.html
ehanley324: I am not sure that the note is necessary.  I struggled with how to invoke it before, let me know if this makes more sense with the algorithm.

index.html
ehanley324: Done.

index.html
ehanley324: Done

index.html
```diff
@@ -196,14 +196,35 @@ <h2><dfn>Performance Timeline</dfn></h2>
...
+      <li>an initially empty <dfn>performance entry buffer map</dfn> <a
+      data-cite="WHATWG-INFRA#ordered-map">map</a>, <a
+      data-cite="WHATWG-INFRA#map-key">keyed</a> on a <code>DOMString</code>,
+      representing the entry type to which the buffer belongs. The <a
+      data-cite="WHATWG-INFRA#ordered-map">map</a>'s <a
+      data-cite="WHATWG-INFRA#map-value">value</a> is the following tuple:
+        <ul>
+          <li>A <dfn>performance entry buffer</dfn> to store
+            <a>PerformanceEntry</a> objects, that is initially empty.</li>
+          <li>A <code>long</code> <dfn>maxBufferSize</dfn> indicating the max
```
npm1: Is 'long' an Infra type? We can just remove that keyword I think.

index.html
```diff
@@ -222,28 +243,29 @@ <h2>Extensions to the <dfn data-cite=
...
-        <h2><dfn>getEntriesByName()</dfn> method</h2>
-        <p>Returns a <a>PerformanceEntryList</a> object returned by <a href=
-        "#filter-buffer-by-name-and-type"></a> algorithm with <var>buffer</var>
-        set to <a>relevant performance entry buffer</a>, <var>name</var> set to
-        `name`, and <var>type</var> set to <code>null</code> if optional
-        `entryType` is omitted, and <var>type</var> set to `type`
-        otherwise.</p>
+        <section>
+          <h2><dfn>getEntries()</dfn> method</h2>
+          <p>Returns a <a>PerformanceEntryList</a> object returned by
```
npm1: nit: 'by the' instead of just 'by'

index.html
```diff
@@ -196,14 +196,35 @@ <h2><dfn>Performance Timeline</dfn></h2>
...
-    <p>The <dfn>relevant performance entry buffer</dfn> is the <a>performance
-    entry buffer</a> associated with the <a data-cite=
-    "HTML/webappapis.html#concept-relevant-global">relevant global
-    object</a>.</p>
+    <p>In order to get the <dfn>relevant performance entry tuple</dfn>, given
+    <var>entryType</var> as input, run the following steps:</p>
+    <ol>
+      <li>Let <var>map</var> be the <a>performance entry buffer map</a> associated
+      with the <a data-cite="HTML/webappapis.html#concept-relevant-global">
+      relevant global object</a> of <a data-cite="DOM#context-object">context
```
npm1: nit: of the context object (add 'the')

index.html
```diff
@@ -222,28 +243,29 @@ <h2>Extensions to the <dfn data-cite=
...
+          <p>Returns a <a>PerformanceEntryList</a> object returned by
+          <a>filter buffer map by name and type</a> algorithm with
+          <var>name</var> and <var>type</var> set to <code>null</code>.</p>
+        </section>
+        <section>
+          <h2><dfn>getEntriesByType()</dfn> method</h2>
+          <p>Returns a <a>PerformanceEntryList</a> object returned by
+          <a>filter buffer map by name and type</a> algorithm with
+          <var>name</var> set to <code>null</code>, and
+          <var>type</var> set to the method's input <code>type</code>
```
npm1: Since you're fixing this, I don't think we need to always specify the name of the parameters when calling an algorithm. For example instead of
algorithm x with name1 set to var1 and name2 set to var2
You can just say something like:
algorithm x passing var1 and var2 as inputs.

The order of the inputs should match the order in the algorithm description.

index.html
```diff
@@ -443,18 +465,15 @@ <h2><dfn>observe()</dfn> method</h2>
...
-                "WHATWG-DOM#context-object">context object</a>'s <a>observer
-                buffer</a>.
-                </li>
-              </ol>
+                <li>
+                  Let <var>tuple</var> be the <a>relevant performance entry
+                  tuple</a> given <var>options</var>'s <a>type</a> as
+                  <var>entryType</var>.</li>
+                <li>For each <a>PerformanceEntry</a> <var>entry</var> in
+                  <var>tuple</var>'s <a>performance entry buffer</a>
```
npm1: nit: slightly better would be
For each entry in tuple's performance entry buffer, append entry to the observer buffer.

index.html
```diff
@@ -660,39 +685,71 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
...
-            <li>Add <var>entryObject</var> to the <var>list of entry objects</var>.
-            </li>
+            <li>Let <var>buffer</var> be <var>tuple</var>'s <a>performance
+            entry buffer</a>.</li>
+            <li>If <var>tuple</var>'s <a>availableFromTimeline</a> is false,
+            continue to the next <var>tuple</var>.</li>
+            <li>Let <var>entries</var> be the result of running <a>filter
+            buffer by name and type</a> with <var>buffer</var>, <var>name</var>
+            and <var>type</var> as inputs.</li>
+            <li>For each <a>PerformanceEntry</a> <var>entry</var> in
```
npm1: nit: similarly wording can be improved here

index.html
```diff
@@ -660,39 +685,71 @@ <h2>Queue a <code>PerformanceEntry</code></h2>
...
+            data-cite="WHATWG-INFRA#list-append">append</a> it to
+            <var>result</var>.</li>
           </ol>
         </li>
-        <li>Return the <var>list of entry objects</var>.
+        <li>Sort <var>results</var>'s entries in chronological order with
+        respect to <a>startTime</a></li>
+        <li>Return <var>result</var>.</li>
+      </ol>
+      <p>When asked to run the <dfn>filter buffer by name and type</dfn>
```
npm1: nit: this algorithm should go in a new section. Alternatively, rename this section with a name that is not equal to the name of the first algorithm in this section.

index.html
```diff
@@ -703,6 +760,19 @@ <h2>Register performance entry type</h2>
         <li>Append <var>type</var> to <a>supported entry types</a>.</li>
       </ol>
     </section>
+    <section data-link-for="PerformanceObserver">
+      <h2>Eligibility for adding a PerformanceEntry to a buffer</h2>
```
npm1: Not sure we need to abstract this mini-algorithm out: it is small and simple and only called once. I would just do the steps inline in the algorithm where it is invoked. No strong opinion though...

npm1:
LGTM from my perspective modulo the nit comments.

index.html
yoavweiss: The reason for splitting this algorithm is clarity. Emily wanted to clarify the reasoning for running these steps, and splitting out that algorithm does that.

index.html
yoavweiss: While that's not a requirement, I think it improves clarity, so maybe leave that in?

yoavweiss:
Thanks for pushing this through!! :)

PR #138, opened 2019-06-27
Remove remaining `buffered` flag

saschanaz:
#114 removed it but one thing has remained.


saschanaz:
Marked as non substantive for IPR from ash-nazg.

PR #143, opened 2019-07-09
Add a basic explainer

npm1:
Currently, there is no explainer. This PR adds an explainer so that developers don't have to read the spec.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,12 +46,13 @@ function getFirstContentfulPaint() {
 ### PerformanceObserver
 
 A PerformanceObserver object can notified of new PerformanceEntry objects, according to their `entryType` value.
-The construction of the object must receive a callback, which will be ran whenever the user agent is dispatching new entries whose `entryType` value match one of the ones being observed by the observer.
+The constructor of the object must receive a callback, which will be ran whenever the user agent is dispatching new entries whose `entryType` value match one of the ones being observed by the observer.
 This callback is not run once per PerformanceEntry nor immediately upon creation of a PerformanceEntry.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,9 +1,139 @@
...
+  // Otherwise, the entry is not there, so return null.
+  else
+    return null;
+}
+```
+
+### PerformanceObserver
+
+A PerformanceObserver object can notified of new PerformanceEntry objects, according to their `entryType` value.
+The construction of the object must receive a callback, which will be ran whenever the user agent is dispatching new entries whose `entryType` value match one of the ones being observed by the observer.
```
yoavweiss: s/construction/constructor/

README.md
```diff
@@ -1,9 +1,139 @@
...
+This callback is not run once per PerformanceEntry nor immediately upon creation of a PerformanceEntry.
+Instead, entries are 'queued' at the PerformanceObserver, and the user agent can execute the callback later.
+When the callback is executed, all queued entries are passed onto the function, and the queue for the PerformanceObserver is reset.
+The PerformanceObserver initially does not observer anything: the `observe()` method must be called to specify what kind of PerformanceEntry objects are to be observed.
+There are two ways to call the `observe()` method, and those should not be mixed within the same PerformanceObserver (an exception will be thrown if this is attempted).
```
yoavweiss: That sentence is not very clear. Maybe "The `observe()` method has two different modes where it can be called with either an entryTypes array or with a single type, as detailed below. Those modes cannot be mixed, or an exception will be thrown"?

PR #148, opened 2019-08-15
Add buffered flag explanation in Introduction

npm1:
This change moves the example description so it occurs before the example. It also adds description of the buffered flag.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/pull/148.html" title="Last updated on Aug 19, 2019, 4:19 PM UTC (58270db)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/148/a07b098...58270db.html" title="Last updated on Aug 19, 2019, 4:19 PM UTC (58270db)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -112,7 +112,21 @@
     <p>Alternatively, the developer can observe the <a>Performance Timeline</a>
     and be notified of new performance metrics and, optionally, previously
     buffered performance metrics of specified type, via the
-    <a>PerformanceObserver</a> interface:</p>
+    <a>PerformanceObserver</a> interface.</p>
+    <p>The <a>PerformanceObserver</a> interface was added in Performance
+    Timeline Level 2 and is designed to address limitations of the buffer-based
+    approach shown in the first example. By using the PerformanceObserver
+    interface, the application can:</p>
+    <ul>
+      <li>Avoid polling the timeline to detect new metrics</li>
+      <li>Eliminate costly deduplication logic to identify new metrics</li>
+      <li>Eliminate race conditions with other consumers that may want to
+      manipulate the buffer</li>
+    </ul>
+    <p>The developer is encouraged to use <a>PerformanceObserver</a> where
+    possible. Further, new performance API's and metrics may only be available
+    through the <a>PerformanceObserver</a> interface. The observer works by specifying a callback in the constructor and specifying the performance entries it's interested in via the <a data-link-for=PerformanceObserver>observe()</a> method. The user agent chooses when to execute the callback, which receives performance entries that have been queued.</p>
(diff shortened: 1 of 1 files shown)


npm1:
PTAL! I will do the column wrapping after I've addressed feedback (to avoid doing it multiple times).

npm1:
I noticed Yoav is OOO, so adding the editor for review :)

index.html
```diff
@@ -112,7 +112,21 @@ <h2>Introduction</h2>
...
+      manipulate the buffer</li>
+    </ul>
+    <p>The developer is encouraged to use <a>PerformanceObserver</a> where
+    possible. Further, new performance API's and metrics may only be available
+    through the <a>PerformanceObserver</a> interface. The observer works by specifying a callback in the constructor and specifying the performance entries it's interested in via the <a data-link-for=PerformanceObserver>observe()</a> method. The user agent chooses when to execute the callback, which receives performance entries that have been queued.</p>
+    <p>To avoid losing entries that are created before the observer is registered for observation, the <a data-link-for=PerformanceObserverInit>buffered</a> flag can be used. When this flag is set, the user agent dispatches records that it has buffered prior to the <a>PerformanceObserver</a>'s creation, and thus they are received in the first callback after the <a data-link-for=PerformanceObserver>observe()</a> call occurs. This enables web developers to register PerformanceObservers when it is convenient to do so without missing out on entries dispatched early on during the page load.</p>
```
igrigorik: > without missing out on entries dispatched early on during the page load.

As written, I don't think this bit is clear and makes it sound like we'll buffer any and all requests. I would flip this framing and lead with challenge of registration during page load. Riffing..

Initial page load creates special considerations when using `PerformanceObserver` interface: a registration must be active to receive events but the registration script may not be available or may not be desired in the critical path. To address this, individual metric providers can choose to buffer some number of events while the page is being constructed, which can be accessed via the <a data-link-for=PerformanceObserverInit>buffered</a> flag when registering the observer. When this flag is set, the user agent retrieves and dispatches events that it has buffered, for specified entry type, and delivers them in the first callback after the <a data-link-for=PerformanceObserver>observe()</a> call occurs. 

Note: The number of buffered events is determined by the specification that defines the metric and buffering is intended to used for first-N events only; buffering is not unbounded or continuous.

WDYT?

npm1: I updated the description, PTAL

npm1:
> Some suggested wording below.
> 
> Also, nitpick: can you please format or run tidy [1] to ensure consistent formatting?
> 
> https://github.com/w3c/performance-timeline/blob/gh-pages/tidyconfig.txt

I mentioned I'd word-wrap after review :) I've run the tidy command and it changed a bunch of the rest of the document as well (no content changes other than the paragraphs in the intro).

PR #154, opened 2019-10-16
chore: fixup some xrefs

### LINKED ISSUE #153: DOMHighResTimeStamp not found in level2
I recently submitted a fix to move the constructor in Level2 because ReSpec would highlight the WebIDL section in red otherwise. But now after that fix I still get a problem about DOMHighResTimeStamp not being found in the PerformanceEntry WebIDL entry. I don't see this problem in gh-pages branch but it is present in level2 branch. @marcoscaceres do you have any idea how I can fix this?

marcoscaceres:
The use of the infra concepts might need some improvement. I'd suggest having a look at the infra spec itself for usage. 

Also, "ECMAScript global environment" is not a defined thing (it's a section in WebIDL, sure... but it's not a defined term)... I assumed you meant the "global object" from HTML? WebIDL defers to HTML for that. Please confirm before merging - if it's not correct, please advise on a more suitable (exported) concept.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/pull/154.html" title="Last updated on Oct 17, 2019, 7:30 AM UTC (23bb645)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/154/714a3cc...23bb645.html" title="Last updated on Oct 17, 2019, 7:30 AM UTC (23bb645)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -43,7 +43,7 @@
         "check-punctuation": true,
       },
       doJsonLd: true,
-      xref: "web-platform",
+      xref: ["hr-time", "hr-time-2", "infra", "html", "dom"],
       mdn: true,
     };
   </script>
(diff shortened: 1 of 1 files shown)


npm1:
Looks good to me, @yoavweiss to confirm that 'global object' is fine here. However, this does not close the mentioned issue, which is related to the level2 branch so please modify the comment so that the issue does not get closed once this lands.

marcoscaceres:
Removed "closes" above.

Issue #163, opened 2020-03-18
Use of "SyntaxError" seems incorrect, should be TypeError


noamr:
In 3.3.1, it is specified that when entryTypes/type is omitted, a SyntaxError should be thrown.
However, SyntaxError are reserved for parsers (like JS syntax, CSS syntax) - not to semantic exceptions such as this one, which should probably be a TypeError.

npm1:
This was added in https://github.com/w3c/performance-timeline/pull/112. @yoavweiss do you recall if we considered TypeError here or not? I think I'd be ok changing it to TypeError.

yoavweiss:
Past me [thought SyntaxError seems like the best fit](https://github.com/w3c/performance-timeline/pull/112#discussion_r241829920).

I believe we were aiming for a "SyntaxError" [DOMException](https://heycam.github.io/webidl/#idl-DOMException-error-names) and not a simple exception, so IIUC it doesn't necessarily refer to parsers. 
I agree it's not ideal, but so does TypeError. A BadInput exception would've been ideal.

noamr:
OK; Maybe then separate PerformanceObserverInit to PerformanceObsrverInitSingle with type+bufferred and PerformanceObserverInitMultiple with entryTypes? That way the IDL-generated code can emit this syntax error automatically... If there is some concensus on this I can suggest a PR

yoavweiss:
That seems like a not-web-exposed change, and could potentially simplify the processing, and enable us to deprecate the "multiple" type in the future.

@npm1 - thoughts?

npm1:
Can you clarify precisely what the proposed IDL change is here? I'm not sure I follow.

noamr:
> Can you clarify precisely what the proposed IDL change is here? I'm not sure I follow.
Sure! something like:
`
dictionary PerformanceObserverInitMultple {
  sequence<DOMString> entryTypes;
};
dictionary PerformanceObserverInitSingle {
  sequence<DOMString> entryTypes;
  DOMString type;
  boolean buffered;
};
...
void observe (optional (PerformanceObserverInitSingle or PerformanceObserverInitMultiple) options = {});
`

npm1:
Oh I see. Assuming we're sure that it's valid IDL, I'm ok with that.

npm1:
[...]

PR #164, opened 2020-04-09
Do not use SyntaxError

npm1:
Fixes https://github.com/w3c/performance-timeline/issues/163

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/pull/164.html" title="Last updated on Apr 9, 2020, 9:37 PM UTC (5fb5262)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/164/87609bc...5fb5262.html" title="Last updated on Apr 9, 2020, 9:37 PM UTC (5fb5262)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -394,10 +394,10 @@
         global object</a>.
         </li>
         <li>If <var>options</var>'s <a>entryTypes</a> and <a>type</a> members
-        are both omitted, then [=exception/throw=] a {{"SyntaxError"}}.
+        are both omitted, then [=exception/throw=] a {{"TypeError"}}.
         </li>
         <li>If <var>options</var>'s <a>entryTypes</a> is present and any other
-        member is also present, then [=exception/throw=] a {{"SyntaxError"}}.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -394,10 +394,10 @@ <h2><dfn>observe()</dfn> method</h2>
         global object</a>.
         </li>
         <li>If <var>options</var>'s <a>entryTypes</a> and <a>type</a> members
-        are both omitted, then [=exception/throw=] a {{"SyntaxError"}}.
+        are both omitted, then [=exception/throw=] a {{"TypeError"}}.
         </li>
         <li>If <var>options</var>'s <a>entryTypes</a> is present and any other
```
npm1, line 399: I noticed that currently Chrome does not throw an error in this case (so for example `observe({entryTypes: ['mark', 'measure'], buffered: true})` will not throw an error, we only issue a console warning. Thoughts?

yoavweiss: Sounds like Chrome should align its implementation. Might be good to add a counter that would help estimate breakage here

PR #180, opened 2021-05-06
Remove context object

npm1:
It seems `context object` is deprecated, in favor of `this`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -411,7 +411,7 @@
       <p>The <a>observe()</a> method instructs the user agent to <dfn>register
       the observer</dfn> and must run these steps:</p>
       <ol data-link-for="PerformanceObserverInit">
-        <li>Let <var>observer</var> be the <a>context object</a>.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -411,7 +411,7 @@ <h2><dfn>observe()</dfn> method</h2>
       <p>The <a>observe()</a> method instructs the user agent to <dfn>register
       the observer</dfn> and must run these steps:</p>
       <ol data-link-for="PerformanceObserverInit">
-        <li>Let <var>observer</var> be the <a>context object</a>.
+        <li>Let <var>observer</var> be <a>this</a>.
```
npm1, line 414: I'm open to removing the variable and replacing all occurrences with `this`, let me know what you think.

yoavweiss: I think it'd be better/clearer to drop the observer var entirely and just use `this`

yoavweiss (approved):
LGTM after dropping the observer var

PR #192, opened 2021-09-15
Add navigationId to PerformanceEntry

clelland:
This adds the concept of a `current navigation id` to the global object, which starts at 0 and is incremented on every navigation event during a page's lifetime. This navigation id then gets exposed on PerformanceEntry objects, to allow those timeline entries to be segmented by user interaction.

***
<a href="https://pr-preview.s3.amazonaws.com/clelland/performance-timeline/pull/192.html" title="Last updated on Nov 26, 2021, 5:37 PM UTC (6e5497e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/192/ca6936d...clelland:6e5497e.html" title="Last updated on Nov 26, 2021, 5:37 PM UTC (6e5497e)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/clelland/performance-timeline/pull/192.html" title="Last updated on Oct 11, 2023, 3:33 PM UTC (1550c7d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/192/3984a34...clelland:1550c7d.html" title="Last updated on Oct 11, 2023, 3:33 PM UTC (1550c7d)">Diff</a>

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -59,6 +59,8 @@ The entry has the following attributes:
 * `entryType`: a string representing the type of performance data being exposed. It is also used to filter entries in the `getEntriesByType()` method and in the `PerformanceObserver`.
 * `startTime`: a timestamp representing the starting point for the performance data being recorded. The semantics of this attribute depend on the `entryType`.
 * `duration`: a time duration representing the duration of the performance data being recorded. The semantics of this one also depend on the `entryType`.
+* `navigationId`: an integer indicating how many times the user has navigated to this document
+since it was initially loaded.
 
 If these sound abstract, it’s because they are.
 A specification whose goal is to expose new measurements to web developers will define a new interface which extends `PerformanceEntry`.
@@ -149,6 +151,21 @@ Having both polling and callbacks has proved to be useful.
 For entryTypes with a fixed amount of entries, like Paint Timing, using the polling method is more convenient, especially in the current state of the world where analytics providers often only send a single beacon per page load.
 At the same time, the callback is more useful for cases where there is varying amount of entries of the given entryType, like in Resource Timing.
 The web performance monitoring service can process all the performance information while the page is still running and only has to do very minimal work when the data needs to be reported.
+
+## Page lifetime issues
+When this API was originally designed, documents had a relatively simple lifecycle: they were
+loaded when the user navigated to them, and unloaded when the user navigated away, with the
+JavaScript environment being torn down at that time. Since then, the sitution has become more
+complex, with many browsers introducing a back-forwards-cache, with which a user can return
+a document wich they have previously navigated away from. The web has also seen a rise in
+popularity of Single Page Apps, where what appears to the user to be a navigation is actually
+just a change in state of a running page. In both of these situations, a navigation (or what
+appears to the user as a navigation) can occur without the performance timeline being reset.
+In order to allow developers to reason about such events during the life of a page,
+PerformanceEntry objects now include a navigation counter field. This field captures the
+count of navigations which had occurred when the entry was generated (starting at 1 when the
+document is initially loaded, and being incremented by one with each subsequent navigation of
+the document.)
 
 # Standards Status
 The Performance Timeline specification is widely approved.
--- a/explainer.md
+++ b/explainer.md
@@ -59,8 +59,7 @@ The entry has the following attributes:
 * `entryType`: a string representing the type of performance data being exposed. It is also used to filter entries in the `getEntriesByType()` method and in the `PerformanceObserver`.
 * `startTime`: a timestamp representing the starting point for the performance data being recorded. The semantics of this attribute depend on the `entryType`.
 * `duration`: a time duration representing the duration of the performance data being recorded. The semantics of this one also depend on the `entryType`.
-* `navigationId`: an integer indicating how many times the user has navigated to this document
-since it was initially loaded.
+* `navigationId`: a string identifying the `PerformanceEntry` object corresponding to the last navigation or navigation-like event that had occurred in the document at the time that this `PerformanceEntry` object was recorded.
 
 If these sound abstract, it’s because they are.
 A specification whose goal is to expose new measurements to web developers will define a new interface which extends `PerformanceEntry`.
@@ -161,11 +160,10 @@ a document wich they have previously navigated away from. The web has also seen 
 popularity of Single Page Apps, where what appears to the user to be a navigation is actually
 just a change in state of a running page. In both of these situations, a navigation (or what
 appears to the user as a navigation) can occur without the performance timeline being reset.
-In order to allow developers to reason about such events during the life of a page,
-PerformanceEntry objects now include a navigation counter field. This field captures the
-count of navigations which had occurred when the entry was generated (starting at 1 when the
-document is initially loaded, and being incremented by one with each subsequent navigation of
-the document.)
+In order to allow developers to reason about such events during the life of a page, some
+PerformanceEntry objects mark navigations, or navigation-like events. All PerformanceEntry
+objects include a navigation ID field, which links each PerformanceEntry to the most recent
+navigation entry which had occurred when the entry was generated.
 
 # Standards Status
 The Performance Timeline specification is widely approved.
--- a/explainer.md
+++ b/explainer.md
@@ -55,11 +55,12 @@ In order to prevent the user agent from storing too much performance data in mem
 This specification defines the `PerformanceEntry` interface, which is used to host performance data of the web application.
 A single `PerformanceEntry` object corresponds to one nugget of information about the performance of the website.
 The entry has the following attributes:
+* `id`: an integer identifying this `PerformanceEntry` object.
 * `name`: a string identifier for the object, also used to filter entries in the `getEntriesByName()` method.
 * `entryType`: a string representing the type of performance data being exposed. It is also used to filter entries in the `getEntriesByType()` method and in the `PerformanceObserver`.
 * `startTime`: a timestamp representing the starting point for the performance data being recorded. The semantics of this attribute depend on the `entryType`.
 * `duration`: a time duration representing the duration of the performance data being recorded. The semantics of this one also depend on the `entryType`.
-* `navigationId`: a string identifying the `PerformanceEntry` object corresponding to the last navigation or navigation-like event that had occurred in the document at the time that this `PerformanceEntry` object was recorded.
+* `navigationId`: an integer identifying (by `id`)  the `PerformanceEntry` object corresponding to last navigation or navigation-like event that had occurred in the document at the time that this `PerformanceEntry` object was recorded.
 
 If these sound abstract, it’s because they are.
 A specification whose goal is to expose new measurements to web developers will define a new interface which extends `PerformanceEntry`.
--- a/explainer.md
+++ b/explainer.md
@@ -155,15 +155,15 @@ The web performance monitoring service can process all the performance informati
 When this API was originally designed, documents had a relatively simple lifecycle: they were
 loaded when the user navigated to them, and unloaded when the user navigated away, with the
 JavaScript environment being torn down at that time. Since then, the sitution has become more
-complex, with many browsers introducing a back-forwards-cache, with which a user can return
-a document wich they have previously navigated away from. The web has also seen a rise in
+complex, with many browsers introducing a back-forward cache, with which a user can return to
+a document which they have previously navigated away from. The web has also seen a rise in
 popularity of Single Page Apps, where what appears to the user to be a navigation is actually
 just a change in state of a running page. In both of these situations, a navigation (or what
 appears to the user as a navigation) can occur without the performance timeline being reset.
 In order to allow developers to reason about such events during the life of a page, some
 PerformanceEntry objects mark navigations, or navigation-like events. All PerformanceEntry
-objects include a navigation ID field, which links each PerformanceEntry to the most recent
-navigation entry which had occurred when the entry was generated.
+objects include a navigation ID field, which ties each PerformanceEntry to the most recent
+navigation entry which had occurred before the entry was generated.
 
 # Standards Status
 The Performance Timeline specification is widely approved.
--- a/index.html
+++ b/index.html
@@ -249,7 +249,6 @@
           <li>An integer <dfn>dropped entries count</dfn> that is initially 0.</li>
         </ul>
       </li>
-      <li>An integer <dfn>current navigation id</dfn> that is initially 0.</li>
     </ul>
     <p>In order to get the <dfn>relevant performance entry tuple</dfn>, given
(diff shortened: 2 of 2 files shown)


index.html
```diff
@@ -258,6 +262,17 @@ <h2><dfn>Performance Timeline</dfn></h2>
         map</var>, given <var>entryType</var> as the <a>key</a>.
       </li>
     </ol>
+    <p>The <a
+    href="https://html.spec.whatwg.org/#session-history-document-visibility-change-steps">session
```
npm1, line 266: Does ReSpec not support links to HTML without having to specify the full url? I thought it does now. And even if it does not, please use the multipage version, as this one is pretty unusable.

index.html
```diff
@@ -344,6 +360,12 @@ <h2>The <dfn>PerformanceEntry</dfn> interface</h2>
         duration concept doesn't apply, a performance metric may choose to
         return a `duration` of <code>0</code>.
       </dd>
+      <dt><dfn>navigationId</dfn></dt>
+      <dd>
+        This attribute MUST return the value of the global object's <a>current
```
npm1, line 365: Just return the value it was set to (this is normative, if you want you can add explanation in non normative section)

index.html
yoavweiss: You'd also need to change the HTML comment saying this algorithm is only used by PageVisibility

yoavweiss (commented):
Looks good! a few nits

index.html
clelland: The term is not exported, as far as I can tell, and [ReSpec](https://respec.org/xref/?term=session+history+document+visibility+change+step) cannot find it. (I've tried a few variants, just in case, but can't get a working xref here, or for the "traverse the history" algorithm linked below. Although if we're changing HTML with this to update the comment, then adding 'export' could happen at the same time.)

yoavweiss:
OK, Now I remember... We have an ongoing discussion on https://github.com/w3c/page-visibility/issues/51 on better ways to hook into the HTML algorithm here, so you probably want to follow the path outlined there, rather than link to that HTML algorithm directly. Happy to chat about this.

/cc @mmocny

clelland:
> OK, Now I remember... We have an ongoing discussion on [w3c/page-visibility#51](https://github.com/w3c/page-visibility/issues/51) on better ways to hook into the HTML algorithm here, so you probably want to follow the path outlined there, rather than link to that HTML algorithm directly. Happy to chat about this.
> 
> /cc @mmocny

That approach makes it look like that the global's navigation id should be entirely defined in HTML; there would seem to be little benefit to declaring it here and then making HTML call an algorithm here just to increment it.

clelland:
Updated to move the navigation counter into HTML (Personal repo commit [here](https://github.com/clelland/html/commit/c09f070cd273b5dc6d7a4df7013c748e2538f0c7)) and use that instead of the *session history document visibility change steps*

Note that this will always be 0 for workers without a relevant document, and should start at 1 for documents with their first load.

yoavweiss:
This LGTM. Let's land the HTML parts first though before landing this.

index.html
marcoscaceres: The visibility stuff got moved to HTML now... might want to coordinate with @noamr on changes.

noamr: `[=page visibility change steps=]` is exported from HTML.
For mature specs (not sure if this change qualified) it's better to put the actual steps linked from HTML rather than use that hook.

clelland: I'll put together an HTML PR for that then. I'll reach out with questions. Thanks @noamr

yoavweiss:
Looking at the discussions on https://github.com/whatwg/html/pull/7370, it seems like it'd be better if we could reach an agreement to land the HTML related processing there...

As currently (experimentally) implemented in [SPA heuristics](https://docs.google.com/document/d/1eUyQg3YLEmYjrTMC3p-F1MilECwCynhM6WIbIo05SPU/edit#), the counter is not incremented on use of the Navigation API, but only when the heuristics are triggered. As SPA heuristics are not (yet) specified, maybe it's possible to convince the HTML folks to name this a navigation counter even if the only user is currently BFCache navigations?

Alternatively, we could name it whatever they are comfortable with, and rename it once SPA heuristics are mature enough to be specified?

paulirish:
After seeing how navigationId works with bfcache/softnavs, I'm surprised by the 1-indexed _appearance_. 

What's the significance of a navId of 0? 

Would that be like.. Network Error Logging reporting on a failed navigation? (Can't be, as NEL isn't a perftimeline concept?)
How would anything on the perf timeline get a navid of 0?

clelland:
> How would anything on the perf timeline get a navid of 0?

With the text as written, only events on a worker's timeline would have a navigationId of 0. A document's counter starts at 0, but is immediately incremented as the document loads.

That said, https://github.com/WICG/soft-navigations/issues/12 is relevant to this discussion, and we should probably be replacing this with a less-counter-y mechanism in any case.

clelland:
Taking the feedback from https://github.com/WICG/soft-navigations/issues/12 into account, I've updated this PR. `navigationId`  is now a string rather than a counter, and all of the algorithms to manage it are here in Performance Timeline.

I expect that it doesn't need HTML integration support any more; I would instead change Navigation Timing to hook into this now, and call "Queue a navigation PerformanceEntry" rather than "Queue a PerformanceEntry".

index.html
clelland: Resolving; the latest patch doesn't hook into HTML here

clelland:
Q: Should we export "most recent navigation" for situations where a timeline entry might be started before a navigation event, but queued after it finishes? I'm thinking about the case where a resource or event timing entry spans a soft-navigation, and we want it to be attributed to the navigation preceding the soft nav, even if it completes afterwards.

clelland:
Ping @yoavweiss -- can you take a look?

explainer.md
```diff
@@ -150,6 +151,20 @@ For entryTypes with a fixed amount of entries, like Paint Timing, using the poll
 At the same time, the callback is more useful for cases where there is varying amount of entries of the given entryType, like in Resource Timing.
 The web performance monitoring service can process all the performance information while the page is still running and only has to do very minimal work when the data needs to be reported.
 
+## Page lifetime issues
+When this API was originally designed, documents had a relatively simple lifecycle: they were
+loaded when the user navigated to them, and unloaded when the user navigated away, with the
+JavaScript environment being torn down at that time. Since then, the sitution has become more
+complex, with many browsers introducing a back-forwards-cache, with which a user can return
```
yoavweiss, line 158: Nit: back-forward cache

explainer.md
```diff
@@ -150,6 +151,20 @@ For entryTypes with a fixed amount of entries, like Paint Timing, using the poll
...
+a document wich they have previously navigated away from. The web has also seen a rise in
```
yoavweiss, line 159: s/wich/which/

explainer.md
```diff
@@ -150,6 +151,20 @@ For entryTypes with a fixed amount of entries, like Paint Timing, using the poll
...
+popularity of Single Page Apps, where what appears to the user to be a navigation is actually
+just a change in state of a running page. In both of these situations, a navigation (or what
+appears to the user as a navigation) can occur without the performance timeline being reset.
```
yoavweiss, line 162: Should we refer to a "same-document navigation" here?

explainer.md
```diff
@@ -150,6 +151,20 @@ For entryTypes with a fixed amount of entries, like Paint Timing, using the poll
...
+In order to allow developers to reason about such events during the life of a page, some
+PerformanceEntry objects mark navigations, or navigation-like events. All PerformanceEntry
+objects include a navigation ID field, which links each PerformanceEntry to the most recent
```
yoavweiss, line 165: s/links/ties/

explainer.md
```diff
@@ -150,6 +151,20 @@ For entryTypes with a fixed amount of entries, like Paint Timing, using the poll
...
+navigation entry which had occurred when the entry was generated.
```
yoavweiss, line 166: s/when/before/ ?

index.html
```diff
@@ -244,6 +247,12 @@ <h2><dfn>Performance Timeline</dfn></h2>
         </ul>
       </li>
     </ul>
+    <p>Each <a>Document</a> has:</p>
+    <ul>
+      <li>a <dfn>most recent navigation</dfn>, which is a <a>PerformanceEntry</a>,
```
yoavweiss, line 252: s/a/A/

index.html
```diff
@@ -883,6 +941,18 @@ <h2>Determine if a performance entry buffer is full</h2>
...
     </section>
+    <section data-link-for="PerformanceEntry">
+      <h2>Generate a navigationId</h2>
+      <p>When asked to <dfn class="export">generate a navigationId</dfn> for a
+      <a>PerformanceEntry</a> <var>entry</a>, run the following steps:</p>
+      <ol>
+        <li>Let <var>type</var> be <var>entry</var>'s <a>entryType</a>.</li>
+        <li>Let <var>startTime</var> be <var>entry</var>'s <a>startTime</a>.
+        </li>
+        <li>Return the [=string/concatenation=] of « <var>type</var>,
```
yoavweiss, line 952: I know we initially talked about a UUID, but this may be equivalent. A few things that may hamper that:
* Can we have multiple navigations per ms? Unlikely, but I'm not sure it's currently prohibited.
* Having a predictable structure may cause developers to create weird reliances on the ID's structure (but I'm not sure that would be the case)
WDYT?

index.html
```diff
@@ -883,6 +941,18 @@ <h2>Determine if a performance entry buffer is full</h2>
...
+        <var>time</var> » using U+002D HYPHEN-MINUS.</li>
```
yoavweiss, line 953: concatenation is defined for a list of strings, so I think you need to convert the timestamp to a string first.

index.html
```diff
@@ -592,24 +607,34 @@ <h2><dfn>PerformanceObserverEntryList</dfn> interface</h2>
...
           <h2><dfn>getEntriesByName()</dfn> method</h2>
           <p>Returns a <a>PerformanceEntryList</a> object returned by <a>filter
-          buffer by name and type</a> algorithm with <a>this</a>'s <a>entry list</a>,
+          buffer</a> algorithm with <a>this</a>'s <a>entry list</a>,
           <var>name</var> set to the method input <code>name</code> parameter, and
           <var>type</var> set to <code>null</code> if optional `entryType` is omitted,
           or set to the method's input <code>type</code> parameter otherwise.</p>
         </section>
+        <section>
+          <h2><dfn>getEntriesByNavigationId()</dfn> method</h2>
```
yoavweiss, line 631: Did we discuss creation of a dedicated filter, compared to applying a JS filter on `getEntries()`? I'm guessing this would have performance benefits, but it's worthwhile to articulate the reasoning (maybe in the explainer)

yoavweiss (commented):
Apologies for the delay! This fell through the cracks..

index.html
noamr: - We can't have two navigations at the same monotonic time.
- I think it's OK that people rely on this if it's well defined. It's only the URL and the start time which are anyway available.

yoavweiss: Makes sense!

FWIW, when working on resource timing initiator (which uses a similar ID concept), we got [feedback](https://docs.google.com/document/d/1ODMUQP9ua-0plxe0XhDds6aPCe_paZS6Cz1h1wdYiKU/edit#heading=h.6m2qmrja1eyi) that it might be more ergonomic to use an object rather than a concatenated string.

clelland: Do we run into problems with object equality, if we use a structured object? Specifically, are there issues around key order or numeric precision, or just the fact that object equality is often conflated with object identity?

Krinkle:
RE: Today's W3C WebPerf presentation by @clelland  in which we saw an integer ("counter") and string idenitifier (`{type}-{startTime}`) proposed as value for the `navigationId`.

I believe startTime could theoretically clash in some unlikely pathalogical scenario where time doesn't move or things happen very fast during recovery from a blockage of sorts. Assuming navigationId has to be unique within a realm, and startTime isn't, does that mean the number would be altered or extended in some way in practice in the event of a conflict? If so, it might be worth reserving something in the format so that consumers won't be caught in surprise when an entry's navigationId varies notably from the assumed string format. For example `{type}-{startTime}-{number}` where the latter would be 0 most cases but mediate would-be conflicts.

It's not clear what value `startTime` adds in that case, since it doesn't aid sorting (we'd need `{startTime}-{type}` instead of `{type}-{startTime}` to aid sorting), and is presumably not meant for machine-readable consumption either (the entry objects have `.startTime` for that).

So perhaps `{type}-{number}` would be simpler if we only want to reflect that it's a blackbox string that is neither an array index nor key for approx time order. Or a step further, and plain `{number}`, as blackbox integer, in which case we're back to a setTimeout-like integer/counter 😁

noamr:
> RE: Today's W3C WebPerf presentation by @clelland in which we saw an integer ("counter") and string idenitifier (`{type}-{startTime}`) proposed as value for the `navigationId`.
> 
> I believe startTime could theoretically clash in some unlikely pathalogical scenario where time doesn't move or things happen very fast during recovery from a blockage of sorts. 

It's monotonic time, so it should be unique % coarsening.

explainer.md
clelland: I'm not sure if that would be confusing - not all same-document navigations are "user-perceived navigations" (fragment navs, for instance), and not all user-perceived navigations are same-document navigations (bfcache restores are not)

Was there wording that you had in mind?

index.html
clelland: Discussed at the WG; there probably are issues with identity/equality here, and a string has better comparison behaviour, as long as we ensure it is unique.
Sticking with <type>-<timestamp> will allow this pattern to be reused for other cases where we want to refer to timeline entries as well.

clelland:
Thinking more about this issue this morning, @yoavweiss, @yashjoshi-dotcom and I came to the conclusion that `{type}-{startTime}` probably won't work as a general id mechanism for performance entries. We can't use it for resource timing or event timing, or even paint timing, as it's almost certain to generate collisions.

However, there is a useful mechanism in event timing already, for [generating interaction ids](https://w3c.github.io/event-timing/#user-interaction-value). I think we can reuse that logic here, to assign an integer id to every entry, and then navigationId will just use the id for the relevant navigation entry.

It's monotonically increasing, but doesn't have a predictable start or increment[^1] so you'd have to go out of your way to try to depend on the exact sequence of values. (And it was the potential for those dependencies which was the reason we didn't want a simple counter.)

[^1]: Should the increment be random each time? Or is a small consistent step better? That's a topic for bikeshedding

explainer.md
yoavweiss: nah, let's land as is, and clarify later if confusion arises.

PR #197, opened 2022-01-27
Add some exports used elsewhere

npm1:
This PR adds some export and applies some fixes for current warnings. When digging into https://github.com/WICG/element-timing/issues/67, I notice that some PerformanceTimeline objects are being manually imported. But this includes some IDL objects. I suspect I shouldn't have to do this?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/pull/197.html" title="Last updated on Jan 28, 2022, 2:46 PM UTC (5d5f115)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/197/ef65cde...5d5f115.html" title="Last updated on Jan 28, 2022, 2:46 PM UTC (5d5f115)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -31,7 +31,7 @@
         },
         {
           name: "Jatinder Mann",
-          url: "mailto: jmann@microsoft.com",
+          url: "mailto:jmann@microsoft.com",
           company: "Microsoft Corp.",
           note: "Until November 2014",
         },
--- a/index.html
+++ b/index.html
@@ -296,7 +296,7 @@
     </section>
   </section>
   <section data-dfn-for="PerformanceEntry" data-link-for="PerformanceEntry">
-    <h2>The <dfn class="export">PerformanceEntry</dfn> interface</h2>
+    <h2>The <dfn>PerformanceEntry</dfn> interface</h2>
     <p>The <a>PerformanceEntry</a> interface hosts the performance data of
     various metrics.</p>
     <pre class='idl'>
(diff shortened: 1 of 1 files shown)


npm1:
By the way, I tried removing the manual import for e.g. PerformanceEntry and it does not find the object in the Element Timing spec. I don't think we're supposed to manually export every IDL object we want to use elsewhere, but not sure what's wrong.

marcoscaceres:
> I suspect I shouldn't have to do this?

Yeah, IDL definitions automatically get exported. So, feel free to remove the `class="export"`. 

> By the way, I tried removing the manual import for e.g. PerformanceEntry and it does not find the object in the Element Timing spec. I don't think we're supposed to manually export every IDL object we want to use elsewhere, but not sure what's wrong.

Right. The document needs to first to end up on /TR/, and then every 6 hours. 

I can see that {{PerformanceEntry}} is already in the definitions database:
https://respec.org/xref/?term=PerformanceEntry

Happy to take a look at Element Timing if it's not linking correctly.

npm1:
Thanks for confirming. I have removed the unneeded exports from the PR.

Regarding the Element Timing issue, if I remove this line[1] then bikeshed complains with
```
LINK ERROR: No 'idl-name' refs found for 'PerformanceEntry'.
<a class="n" data-link-type="idl-name" data-lt="PerformanceEntry">PerformanceEntry</a>
```
I'm not sure how to debug that. The link error seems to come from the IDL for PerformanceElementTiming.

[1] https://github.com/WICG/element-timing/blob/main/index.bs#L18

marcoscaceres (approved):

marcoscaceres:
> Regarding the Element Timing issue, if I remove this line[1] then bikeshed complains with...

Yeah, that's weird... you shouldn't need `<pre class=anchors>` at all as specs should be exporting their definitions :(

PR #202, opened 2022-09-21
Add support for accessing cross-frame entries.

clelland:
This adds a Document-Policy configuration point for documents, named "share-performance-timeline-with", which acceps a set of origins. If it is non-empty, then ancestor frames whose origins are contained in that set can access the timeline entries of the child frame, either through a PerformanceObserver, or by passing a boolean flag into the various getEntries methods.

***
<a href="https://pr-preview.s3.amazonaws.com/clelland/performance-timeline/pull/202.html" title="Last updated on Jan 10, 2023, 8:00 PM UTC (2854086)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/202/3984a34...clelland:2854086.html" title="Last updated on Jan 10, 2023, 8:00 PM UTC (2854086)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -263,31 +263,50 @@
       hosts performance related attributes and methods used to retrieve the
       performance metric data from the <a>Performance Timeline</a>.</p>
       <pre class="idl">
+      dictionary PerformanceEntryOptions {
+        DOMString type;
+        DOMString name;
+        boolean includeFrames;
+      };
+
       partial interface Performance {
-        PerformanceEntryList getEntries (optional boolean includeFrames = false);
-        PerformanceEntryList getEntriesByType (DOMString type, optional boolean includeFrames = false);
-        PerformanceEntryList getEntriesByName (DOMString name, optional DOMString type, optional boolean includeFrames = false);
+        PerformanceEntryList getEntries (optional PerformanceEntryOptions options = {});
+        PerformanceEntryList getEntriesByType (DOMString type);
+        PerformanceEntryList getEntriesByName (DOMString name, optional DOMString type);
       };
       typedef sequence&lt;PerformanceEntry&gt; PerformanceEntryList;
       </pre>
       <p>The <dfn>PerformanceEntryList</dfn> represents a sequence of
       <a>PerformanceEntry</a>, providing developers with all the convenience
       methods found on JavaScript arrays.</p>
+      <section data-dfn-for="PerformanceEntryOptions" data-link-for=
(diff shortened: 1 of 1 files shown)


clelland:
I think this is still missing a piece -- PerformanceObservers shouldn't always get events from child frames; that needs to be conditional on PerformanceObserverInit.includeFrames.

Additionally, the way this is written may cause the finite buffers to overflow quickly; I should find a way to avoid that

yoavweiss:
This seems generally reasonable. May be good to share this with the WG, and see what folks think of the general shape

clelland:
WG is generally supportive (discussed at the 2022-11-10 meeting)

One question was brought up about feature detection, and how developers could know whether cross-frame entries are supported at all, if we just use a boolean parameter on `getEntries*`.

After some consideration, I think that developers should be able to detect the presence of this feature by looking for the `source` property on the `PerformanceEntry` interface; something like

```
Object.getOwnPropertyNames(PerformanceEntry.prototype).indexOf('source') != -1
```

index.html
```diff
@@ -263,9 +265,9 @@ <h2>Extensions to the {{Performance}} interface</h2>
       performance metric data from the <a>Performance Timeline</a>.</p>
       <pre class="idl">
       partial interface Performance {
-        PerformanceEntryList getEntries ();
-        PerformanceEntryList getEntriesByType (DOMString type);
-        PerformanceEntryList getEntriesByName (DOMString name, optional DOMString type);
+        PerformanceEntryList getEntries (optional boolean includeFrames = false);
```
noamr, line 268: This and the subsequent calls are a boolean trap. I just ran into this in a test that said `getEntries(true)` and it was awfully confusing.

Should be `getEntries({includeFrames: false})` or so.

noamr (changes requested):
Let's resolve the boolean trap... I think it's material.

index.html
clelland: Updated; I've added an options dict for `getEntries` -- and only for `getEntries`, at that point, `getEntriesByType`/`getEntriesByName` are kind of redundant, so there's no benefit to changing their signatures.

noamr:
Another thing I didn't see here or in the design document is reference to to `timeOrigin`. Entries from different frames are going to be of different time origins and thus out of sync. I guess we could have the time origin as a property in each cross-frame entry? Would that be ergonomic?

Issue #218, opened 2025-03-18
PerformanceEntry.id shadows  PerformanceElementTiming.id.


powdercloud:
`PerformanceEntry` is defined here:
https://w3c.github.io/performance-timeline/#the-performanceentry-interface

`PerformanceElementTiming` is defined here:
https://w3c.github.io/element-timing/#sec-performance-element-timing

`PerformanceElementTiming` extends `PerformanceEntry` -  yet both define an id attribute, with different types and different meaning.

```
interface PerformanceEntry {
  readonly    attribute unsigned long long  id;
...
```

```
interface PerformanceElementTiming : PerformanceEntry {
    ...
    readonly attribute [DOMString] id;
...
```

Looking at both specifications, the `id` in `PerformanceElementTiming` appears more useful - it is set to the `id` of the element that was observed; whereas, the `id` in Performance entry is some number that increments between subsequent entries, with no further meaning except that it currently takes part in initializing the `PerformanceEntry.navigatinId` (see Step 5 of https://w3c.github.io/performance-timeline/#queue-a-performanceentry).

Therefore, I'm thinking it would be best to remove `PerformanceEntry.id`, and slightly rephrase the mechanism for assigning `PerformanceEntry.navigationId` so that it no longer references that attribute.

Please let me know what you think.

npm1:
Yea incorrectly added in https://github.com/w3c/performance-timeline/pull/192. I don't think Chrome shipped the PerformanceEntry `id`, it should be removed from the spec. https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/timing/performance_entry.idl

mmocny:
Thanks @npm1 for chiming in!

I wonder if you or @clelland recall: was there an explicit design goal with introducing an `id` type for every PerformanceEntry (distinct from navigationId)?

One reason might be to help with ordering, without relying on startTime.

For example, in Event Timing, it is possible for two sequential events (as per event dispatch / event processing time) to actually not have sequential `timeStamp` values, and so therefore not have a sequential `startTime` values [1].

Because we currently try to keep performance entries sorted by `startTime`, this means reporting isn't actually always in "entry creation" order.  Though probably we could solve that without adding an `id`...

---

[1]: Potentially this could happen if the browser process observes these events in one order, but the hardware timestamps are another order-- especially if using two different input modalities. 
[...]

PR #219, opened 2025-03-18
Remove PerformanceEntry.id and rephrase id generation.

powdercloud:
- Removing PerformanceEntry.id, since it's shadowed by PerformanceElementTiming.id, see https://github.com/w3c/performance-timeline/issues/218
- Rephrasing how navigation ids are generated:
  - no more mention of id, but the mechanism is the same as before, a random navigation id for each new document, which is then incremented for navigations that don't change the document
  - changes the order of the sections for queing so that it first talks about navigation performance entries, since these change the navigation ids, and only after that other entries

***
<a href="https://pr-preview.s3.amazonaws.com/powdercloud/performance-timeline/pull/219.html" title="Last updated on May 30, 2025, 11:38 PM UTC (4524f6c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/performance-timeline/219/7074783...powdercloud:4524f6c.html" title="Last updated on May 30, 2025, 11:38 PM UTC (4524f6c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -119,7 +119,7 @@
     and be notified of new performance metrics and, optionally, previously
     buffered performance metrics of specified type, via the
     <a>PerformanceObserver</a> interface.</p>
-    <p>The <a>PerformanceObserver</a> interface was added 
+    <p>The <a>PerformanceObserver</a> interface was added
     and is designed to address limitations of the buffer-based
     approach shown in the first example. By using the PerformanceObserver
     interface, the application can:</p>
@@ -315,7 +315,7 @@
         readonly    attribute DOMString           entryType;
         readonly    attribute DOMHighResTimeStamp startTime;
         readonly    attribute DOMHighResTimeStamp duration;
-        readonly    attribute long                navigationId;
+        readonly    attribute unsigned long       navigationId;
         [Default] object toJSON();
       };</pre>
     <dl>
--- a/index.html
+++ b/index.html
@@ -247,15 +247,12 @@
           <li>An integer <dfn>dropped entries count</dfn> that is initially 0.</li>
         </ul>
       </li>
-      <li>An integer <dfn>last performance entry id</dfn> that is initially set
-      to a random integer between 100 and 10000.
+      <li>An integer <dfn>current navigation id</dfn>, which is initially 0.
       </li>
     </ul>
     <p>Each <a>Document</a> has:</p>
     <ul>
-      <li>A <dfn>most recent navigation</dfn>, which is a <a>PerformanceEntry</a>,
-      initially unset.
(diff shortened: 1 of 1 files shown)


mmocny (commented):
(Nit: swapping the order of the algorithms, while cleaner, makes it slightly harder to review the changes in isolation.  Would it be possible to keep the order for initial review?)

I think the change generally looks good.  However, looking at the wording, I wonder if we really need a "most recent navigation" value which is set to a specific navigation entry.

This has a few downsides:
- Requires a mostly similar redundant copy of the queue and entry algorithm
- Requires that we increment navigationId only after we finalize the navigation entry.  We might want to increment it before the navigation timing is fully known (in theory).

What if we just have a "current navigationId" which is always default initialized to a random number, always incremented by a small integer, and some other spec calls into a simpler "increment navigation id" algorithm or some such thing?

powdercloud:
I swapped the sections back for now, so it looks that in Github's file view it's now easy to see a diff of the modifications relative to the 'production' version.
I think with the wording, you're right, it's probably just the most recent navigation id that matters here. I think this is also what I'm seeing in the Chromium change that I'm preparing actually, the entry isn't kept around, just the id.
And, there may be one more state that can occur, when there isn't a document yet - in the code I'm using 0 thus far in that case (for 'uninitialized'?), but I don't know yet whether that value ever actually surfaces via the API (so it wouldn't need to be in the spec, just an implementation wrinkle). If it's cool, I'll come back to this spec update in a little while with more knowledge (1-2 days).

powdercloud:
I uploaded a new version, main changes are:
* [In the initial section](https://pr-preview.s3.amazonaws.com/powdercloud/performance-timeline/pull/219.html#performance-timeline), as suggested, specified "current navigation id" and "current document navigation id", both initialized to 0.
* In the [IDL for PerformanceEntry](https://pr-preview.s3.amazonaws.com/powdercloud/performance-timeline/pull/219.html#the-performanceentry-interface), revised the type of navigationId to long (meaning a signed int32). I think this is better for json and javascript APIs; the extra range of unsigned long long isn't needed and may cause clipping / loss of precision (e.g. the 54 bit limit) and/or things getting emitted as floating point numbers with a decimal dot.
* Revised [Generate a navigation id](https://pr-preview.s3.amazonaws.com/powdercloud/performance-timeline/pull/219.html#generate-a-navigation-id) to use current navigation id and current document navigation id.
* Merged the case of queing a navigation performance entry into [Queue a PerformanceEntry](https://pr-preview.s3.amazonaws.com/powdercloud/performance-timeline/pull/219.html#queue-a-performanceentry).

